PR #16, opened 2018-08-18
Refactor + Exit NFT

jdkanani:
- [x] Use open zeppelin for ERC contracts
- [x] Break down RootChain contract into pieces for better readability
- [x] Add NFT for the exit for plasma. Create NFT on exit and burn on actual withdraw or on the challenge
- [x] Add burn withdraw for reusable account
- [x] Exit without burn (this will make account un-usable)
- [ ] Add deposit withdraw
- [ ] Add challenges for all types of exits
- [ ] Add test-cases for each case


jdkanani:
@ashishrp @shogochiai @vaibhavchellani would love your review. It's WIP!

contracts/lib/Common.sol
```diff
@@ -32,7 +32,16 @@ library Common {
...
+  function toUint8(bytes _arg) internal view returns (uint8) {
     return uint8(_arg[0]);
   }
+
+  /**
+  * @dev Returns the maximum of two numbers
+  * @param a uint256 number
+  * @param b uint256 number
+  */
+  function max(uint256 a, uint256 b) internal pure returns (uint256) {
```
syuhei176: why not use library?
https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/math/Math.sol#L9

jdkanani: Yeah I will add. Missed it.

jdkanani:
Added `LogDecoder` to decode nested logs 

For more info, check https://github.com/trufflesuite/truffle/issues/555

contracts/lib/Common.sol
jdkanani: Added in new commit https://github.com/maticnetwork/contracts/pull/16/commits/9d65877b2d7c1e7744244c9b01a3288291694f70

contracts/root/WithdrawManager.sol
```diff
@@ -0,0 +1,484 @@
...
+      txRoot,
+
+      rootToken,
+      receiptAmount,
+
+      msg.sender
+    );
+
+    // withdraw
+    _withdraw(
```
shogochiai: 1. This `_withdraw()` function is called by both burnProofWithdrawal-logic and exitWithdrawal-logic
2. This function appends withdrawal to priorityQueue finally
3. burnProofWithdrawal must call `WETH.withdrawal` or `ERC20.transfer` rather than `_withdraw`

Is this statement correct? @jdkanani

contracts/root/WithdrawManager.sol
```diff
@@ -0,0 +1,484 @@
...
+
+      path,
+      0,
+
+      msg.sender
+    );
+  }
+
+  // withdraw tokens
+  function withdrawTokens(
```
shogochiai: This function was used in withdrawal-bridge, but the w-bridger is deprecated. So this function must be removed for the sake of readability.

contracts/root/WithdrawManager.sol
```diff
@@ -0,0 +1,484 @@
...
+      }
+
+      // get withdraw block
+      PlasmaExit memory currentExit = exits[utxoPos];
+
+      // process if NFT exists
+      // If an exit was successfully challenged, owner would be address(0).
+      address exitOwner = ExitNFT(exitNFTContract).ownerOf(utxoPos);
+      if (exitOwner != address(0)) {
+        if (_token == wethToken) {
```
shogochiai: This condition scope had better to be separated function because possibly called from `withdrawal()` function.

contracts/root/WithdrawManager.sol
jdkanani: No. `burnProofWithdrawal` will also call `_withdraw()` as operator can just create coins out of thin air, then burn it and create checkpoint without providing data - data unavailability issue.

So, if you burn then you can reuse your account. But if you don't burn and start exit without it, anyone can come and challenge if you reuse your account.

Reuse account == make transaction on Matic chain after exit

contracts/root/WithdrawManager.sol
jdkanani: This function is to withdraw without burn (last resort). When user is done with Matic and Matic chain is no longer functional (extreme case)

We will update `withdrawal-bridge` once this pull request is done.

contracts/root/WithdrawManager.sol
jdkanani: I didn't get what you saying.

contracts/root/WithdrawManager.sol
shogochiai: make sense :)

contracts/root/WithdrawManager.sol
shogochiai: 1. 
> as operator can just create coins out of thin air, then burn it and create checkpoint without providing data - data unavailability issue.

make sense :)

2. 
> if you don't burn and start exit without it, anyone can come and challenge if you reuse your account.

Additionally anyone can cause state transition (e.g. balance addition) and the exit will be malicious.
So this exit function must be either "burnt exit" or "last resort"

contracts/root/WithdrawManager.sol
```diff
@@ -147,7 +147,7 @@ contract WithdrawManager is DepositManager {
   }
 
   // withdraw tokens
-  function withdraw(
+  function withdrawBurntTokens(
```
shogochiai: Questions and Assumptions:

Q. How long does it take until child-chain burnProof confirmed?
Asm. `~=3600sec`, until committed to rootchain

Q. How long does it take until `withdrawBurntTokens` got processed?
Asm. `Math.max256(_createdAt + 2 weeks, block.timestamp + 1 weeks)`

I'm bit missing spec. Plz feed me back when you are somehow free @jdkanani

contracts/root/WithdrawManager.sol
shogochiai: never mind, question solved by other answer.

contracts/root/WithdrawManager.sol
jdkanani: Yes!

jdkanani:
@syuhei176 can you take a look at the structure? I will create a diagram to explain in more details.

jdkanani:
This pull request is getting really getting big. We will merge this and keep working on other issues.

vaibhavchellani:
yes makes sense

PR #21, opened 2018-09-28
Stake manager

0xAshish:
simple queue-based stake manager with the random selection of validators, here we intend to keep stake stable/equal for everyone.


jdkanani:
@0xAshish why closed?

0xAshish:
this one is child branch of PQ based stake manager so it contains all those commit messages, so will cut a branch form matic/master and create a new PR.

PR #38, opened 2018-11-28
Add heimdall changes

jdkanani:
@vaibhavchellani Please review

@0xAshish Also make the user a validator if the spot is open. Currently, it happening from next epoch.

### DIFF
--- a/contracts/root/RootChain.sol
+++ b/contracts/root/RootChain.sol
@@ -26,15 +26,6 @@ contract RootChain is Ownable, IRootChain {
   using RLP for RLP.RLPItem;
   using RLP for RLP.Iterator;
 
-  // header block
-  // struct HeaderBlock {
-  //   bytes32 root;
-  //   uint256 start;
-  //   uint256 end;
-  //   uint256 createdAt;
-  //   address proposer;
-  // }
-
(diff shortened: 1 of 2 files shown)


contracts/root/RootChain.sol
```diff
@@ -117,30 +108,28 @@ contract RootChain is Ownable, IRootChain {
...
     // validate extra data using getSha256(extradata)
-    require(keccak256(dataList[6].toData()) == keccak256(bytes20(sha256(extradata))));
+    require(keccak256(dataList[5].toData()) == keccak256(bytes20(sha256(extradata))), "Extra data is invalid");
 
     // extract end and assign to current child
-    dataList = extradata.toRLPItem().toList()[0].toList();
-    uint256 start = currentChildBlock();
-    uint256 end = dataList[2].toUint();
-    bytes32 root = dataList[3].toBytes32();
+    dataList = extradata.toRLPItem().toList();
```
vaibhavchellani: does this work ? Dont we need to do like `extradata.toRLPItem().toList()[0].toList();` ?

jdkanani: Yeah, it will work as I am only decoding msg data not basetx.

PR #39, opened 2018-11-29
test case for submit headerblock

0xAshish:


jdkanani:
@0xAshish check #38 and add test cases according to that.

0xAshish:
> @0xAshish check #38 and add test cases according to that. You can push test-cases in same pull request.

Sure.

PR #50, opened 2019-02-05
Tokenized pos

0xAshish:


contracts/root/StakeManager.sol
```diff
@@ -37,118 +30,134 @@ contract StakeManager is StakeManagerInterface, RootChainable, Lockable {
...
   }
 
-  function stakeFor(address user, address unstakeValidator, address signer, uint256 amount) public onlyWhenUnlocked {
-    require(stakers[user].epoch == 0, "No second time staking");
-    // currentValidatorSetSize*2 means everyone is commited
-    require(validatorThreshold*2 > validatorList.currentSize(), "Validator set full");
-    require(amount < MAX_UINT96, "Stay realistic!!");
+  function stakeFor(address user, uint256 amount, address signer, uint96 validatorId) public onlyWhenUnlocked {
+    require(validators[user].epoch == 0, "No second time staking");
+    require(validatorId > 0 && validatorId <= validatorThreshold);
```
jdkanani: I was hoping validator Id will be different than total validators. Sometimes we validator exits, it vacates the seat but can still have validatorId till it gets rewards (and during withdraw delay period)

contracts/root/StakeManager.sol
```diff
@@ -1,33 +1,26 @@
...
 import { RootChainable } from "../mixin/RootChainable.sol";
 
-import { StakeManagerInterface } from "./StakeManagerInterface.sol";
-import { RootChain } from "./RootChain.sol";
+import { Validator } from "./Validator.sol";
+import { IStakeManager } from "./IStakeManager.sol";
 
 
-contract StakeManager is StakeManagerInterface, RootChainable, Lockable {
+contract StakeManager is IStakeManager, Validator, RootChainable, Lockable {
```
jdkanani: Why extend `StakeManager` with `Validator` and not use a different contract? StakeManager will be big as there will be reward/slash logic will come too.

contracts/root/StakeManager.sol
```diff
@@ -37,118 +30,134 @@ contract StakeManager is StakeManagerInterface, RootChainable, Lockable {
...
         validatorState[dPlusTwo].amount +
-        int256(amount) - int256(stakers[unstakeValidator].amount)
+        int256(amount) - int256(validators[unstakeValidator].amount)
       );
-      emit UnstakeInit(unstakeValidator, stakers[unstakeValidator].amount, dPlusTwo);
+
+      // _clearApproval(unstakeValidator, validatorId);
+      // _removeTokenFrom(unstakeValidator, validatorId);
+      // _addTokenTo(user, validatorId);
+      _burn(unstakeValidator, validatorId);
```
jdkanani: Check my comment on extending `Validator`

contracts/root/StakeManager.sol
```diff
@@ -261,17 +276,17 @@ contract StakeManager is StakeManagerInterface, RootChainable, Lockable {
...
+      validators[user].activationEpoch <= currentEpoch ) &&
+      (validators[user].deactivationEpoch == 0 ||
+      validators[user].deactivationEpoch > currentEpoch)
     );
   }
 
   function checkSignatures (
     bytes32 voteHash,
-    bytes sigs
+    bytes sigs //, bytes bitArray
```
jdkanani: Remove `byte bitArry` comment

contracts/root/Validator.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.4.24;
+
+import { ERC721 } from "openzeppelin-solidity/contracts/token/ERC721/ERC721.sol";
+
+contract Validator is ERC721 {
+
```
jdkanani: Why not use as different contract instead of extending to StakeManager?

contracts/root/StakeManager.sol
jdkanani: In my view, `validatorId` is not slot position.

contracts/root/Validator.sol
0xAshish: yeah, I was kinda thinking like we can use it as storage, push all properties to the validator contract and in stake manager use it on an abstract level(like primitive data type).

contracts/root/StakeManager.sol
0xAshish: IMO slashing, rewards should be done on a separate contract which communicates with root chain and stage manager both, also thinking about abstracting out the validator data(as our discussion in other comments).

contracts/root/IStakeManager.sol
```diff
@@ -2,13 +2,13 @@ pragma solidity ^0.4.24;
...
-interface StakeManagerInterface {
-  event Staked(address indexed user, address indexed signer, uint256 indexed activatonEpoch, uint256 amount, uint256 total);
+interface IStakeManager {
+  event Staked(address indexed user, uint256 indexed activatonEpoch, uint256 amount, uint256 total);
   event Unstaked(address indexed user, uint256 amount, uint256 total);
 
-  function stake(address unstakeValidator, address signer, uint256 amount) public;
-  function stakeFor(address user, address unstakeValidator, address signer, uint256 amount) public;
-  function unstake() public;
+  function stake(uint256 amount, address signer, uint96 validatorId) public;
```
jdkanani: should we need `validatorId`?

contracts/root/IStakeManager.sol
```diff
@@ -2,13 +2,13 @@ pragma solidity ^0.4.24;
...
+  function stakeFor(address user, uint256 amount, address signer, uint96 validatorId) public;
```
jdkanani: should we need `validatorId`?

contracts/root/IStakeManager.sol
```diff
@@ -2,13 +2,13 @@ pragma solidity ^0.4.24;
...
+  function unstake(uint96 validatorId) public;
```
jdkanani: should we need `validatorId`?

contracts/root/StakeManager.sol
```diff
@@ -1,201 +1,172 @@
...
+  mapping (address => uint256) public signerToValidator;
+  // validator metadata
+  mapping (uint256 => Validator) public validators;
   //Mapping for epoch to totalStake for that epoch
   mapping (uint256 => State) public validatorState;
 
-  constructor () public {
-    validatorList = new AvlTree(); // TODO: bind with stakemanager
-  }
+  constructor () ERC721Full("StakeManager", "SM") public {}
```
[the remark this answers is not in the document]
0xAshish: yeah, that sounds good.

contracts/root/IStakeManager.sol
0xAshish: it was an old design, now removed.

contracts/root/StakeManager.sol
0xAshish: now every validator has unique ValidatorId so no conflicts while syncing etc.

jdkanani:
Looks good to me.

PR #56, opened 2019-03-08
 minor refactoring and potential gas savings!

atvanguard:
Some refactoring to make the code more idiomatic to start warming up to the code.

This PR is based on top of https://github.com/maticnetwork/contracts/pull/55

### DIFF
--- a/contracts/proofs/ExitValidator.sol
+++ b/contracts/proofs/ExitValidator.sol
@@ -19,7 +19,7 @@ contract ExitValidator is RootChainValidator {
 
     bytes txBytes,
     bytes txProof
-  ) external {
+  ) public {
     validateExitId(exitId, txBytes);
 
     // validate tx
--- a/contracts/proofs/ExitValidator.sol
+++ b/contracts/proofs/ExitValidator.sol
@@ -3,7 +3,6 @@ pragma solidity ^0.4.24;
 import { RootChainValidator } from "../mixin/RootChainValidator.sol";
 import { RootChain } from "../root/RootChain.sol";
 
-
 contract ExitValidator is RootChainValidator {
   // challenge exit
   function challengeExit(
@@ -20,9 +19,8 @@ contract ExitValidator is RootChainValidator {
 
     bytes txBytes,
     bytes txProof
-  ) public {
-    // validate exit id and get owner
-    address owner = validateExitId(exitId);
+  ) external {
+    validateExitId(exitId, txBytes);
 
     // validate tx
     require(
@@ -40,9 +38,6 @@ contract ExitValidator is RootChainValidator {
         txProof
       )
     );
-
-    // check if owner is same
-    require(getTxSender(txBytes) == owner);
 
     // check if tx happened after exit
     uint256 exitId2 = (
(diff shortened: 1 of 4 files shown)


contracts/proofs/ExitValidator.sol
```diff
@@ -20,9 +19,8 @@ contract ExitValidator is RootChainValidator {
 
     bytes txBytes,
     bytes txProof
-  ) public {
-    // validate exit id and get owner
-    address owner = validateExitId(exitId);
+  ) external {
```
atvanguard: `External functions are sometimes more efficient when they receive large arrays of data.` For public functions these arrays need to be copied to be memory whereas external functions can read directly from calldata. Since this function has several bytes arrays, making it external might make sense.

contracts/root/ExitManager.sol
```diff
@@ -71,24 +71,6 @@ contract ExitManager is RootChainable {
   // Public functions
   //
 
-  /**
-  * @dev Returns information about an exit.
-  * @param _utxoPos Position of the UTXO in the chain.
-  * @return A tuple representing the active exit for the given UTXO.
-  */
-  function getExit(uint256 _utxoPos)
```
atvanguard: This function is redundant since exits mapping is public. See corresponding client contract changes above. Less code is gas savings in deployment.

contracts/proofs/ExitValidator.sol
atvanguard: Interestingly, making it external fails with `/contracts/contracts/proofs/ExitValidator.sol:53:37: CompilerError: Stack too deep, try removing local variables.` - when actually it should be using lesser stack memory! Will dig in.

contracts/root/ExitManager.sol
0xAshish: We have added this function so that users/clients don't have to look up to the contracts to know what is being returned as we cant return the whole struct. [See More](https://github.com/ethereum/web3.js/issues/1728)

0xAshish:
@atvanguard can you please remove refactoring changes and keep those `npm i` changes so that we can merge that PR, and you can open another PR for refactoring and figure out stuff?

atvanguard:
Yes, that is this (PR)[https://github.com/maticnetwork/contracts/pull/55]

contracts/root/ExitManager.sol
atvanguard: How is lookup a problem?
`ExitManager.methods.exits(1).call();` will return JSON data of the form
```
{ '0': '0x4aA2032526286188d7ef812c2d4702E0AeffF0ad',
  '1': '0x0000000000000000000000000000000000000001',
  '2': '100',
  '3': true,
  owner: '0x4aA2032526286188d7ef812c2d4702E0AeffF0ad',
  token: '0x0000000000000000000000000000000000000001',
  amountOrTokenId: '100',
  burnt: true }
```

contracts/proofs/ExitValidator.sol
atvanguard: Apparently, bytes types in calldata need two stack slots, while they only need a single stack slot in memory. See (this)[https://github.com/ethereum/solidity/issues/5751]

atvanguard:
Change will be obsolete, because we are planning a major contracts refactor.

PR #57, opened 2019-03-22
correct ordering of event params name

0xAshish:


0xAshish:
made changes in the new refactoring branch so closing it now.

PR #58, opened 2019-03-28
Migrations to deploy contracts on child chain, add and map token

atvanguard:


package.json
```diff
@@ -50,14 +56,14 @@
...
     "ethereumjs-wallet": "0.6.2",
     "ethers": "4.0.13",
+    "pre-commit": "^1.2.2",
     "safe-buffer": "5.1.2",
     "solium": "^1.1.8",
-    "truffle": "4.1.14",
-    "web3": "^1.0.0-beta.30",
-    "pre-commit": "^1.2.2",
-    "eth-sig-util": "^2.1.1"
+    "truffle": "^5.0.10",
```
jdkanani: @atvanguard  Did we also upgrade zeppelin?

atvanguard: No, why is that required?

atvanguard:
Making a new [PR](https://github.com/maticnetwork/contracts/pull/59) on master branch instead.

package.json
jdkanani: New zeppelin has some functions we can use. @0xAshish has put some hacks around it. Now that we are upgrading things, upgrading zeppeling would make sense.

atvanguard: Great! @0xAshish keep an eye out for any such opportunity to incorporate the features from zep.

PR #66, opened 2019-04-10
add child chain contracts

0xAshish:


atvanguard:
Should be ok to open this PR against, `dev-refactoring`

contracts/child/ChildToken.sol
```diff
@@ -36,10 +36,10 @@ contract ChildToken is Ownable {
     uint256 output1
   );
 
-  function deposit(address user, uint256 amountOrTokenId) public onlyOwner;
+  function deposit(address user, uint256 amountOrTokenId) public;
 
   function withdraw(uint256 amountOrTokenId) public;
 
-  function setParent(address parent) public isParentOwner;
+  function setParent(address _parent) public;
```
atvanguard: Since this contract `is Ownable`, the `parentOwner` variable can be removed.
What does the `parent` of the token contract mean? and why did you remove the `isParentOwner` modifier from `setParent`?

contracts/child/IParentToken.sol
```diff
@@ -1,7 +1,7 @@
-pragma solidity ^0.4.24;
+pragma solidity ^0.5.2;
 //interface for parent contract of any child token
 
 
-interface IParentToken {
+contract IParentToken {
```
atvanguard: How about we let this be an interface and make `beforeTransfer` external.

contracts/child/ParentTokenMock.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity ^0.4.24;
+pragma solidity ^0.5.2;
```
atvanguard: Let's remove this mock. I trying to remove all mocks from the code. It isn't making any sense to have contract mocks.

0xAshish: this is an example contract for testing our pre-transfer hooks.

contracts/child/IParentToken.sol
0xAshish: beforeTransfer won't be called externally 😆.

contracts/child/ChildToken.sol
0xAshish: because abstract functions can't have modifiers.
`parent ` is for before transfer contract and `parentOwner ` is its owner.

contracts/child/ParentTokenMock.sol
0xAshish: this is just an example contract we are using to test beforeTransfer function.

contracts/child/IParentToken.sol
0xAshish: can't, `beforeTransfer ` will never be called externally.

atvanguard: It will be, but let's ignore it for now :)

contracts/child/ChildToken.sol
atvanguard: As discussed, `parentOwner` design is a little twisted, but ok for now.

contracts/child/IParentToken.sol
0xAshish: reverted back to interface.

PR #68, opened 2019-04-17
Add stakemanager fastexit

0xAshish:


contracts/root/RootChain.sol
```diff
@@ -68,6 +68,20 @@ contract RootChain is RootChainStorage, IRootChain {
     _blockDepositId.add(1);
   }
 
+  function getHeaderBlock(uint256 _headerNumber) external view returns (
```
atvanguard: Please remove this function. `headerBlocks` mapping public.

contracts/common/mixin/RootChainValidator.sol
```diff
@@ -1,95 +1,78 @@
...
 
     // check if tx's block is included in header and tx is in block
-    return keccak256(blockNumber, blockTime, txRoot, receiptRoot)
+    return keccak256(abi.encodePacked(blockNumber, blockTime, txRoot, receiptRoot))
       .checkMembership(blockNumber - start, headerRoot, headerProof) &&
     MerklePatriciaProof.verify(txBytes, path, txProof, txRoot);
   }
 
   // validate transaction & receipt
   function validateTxReceiptExistence(
```
atvanguard: Let's clean up the functions here. e.g. calling this func `validateReceiptExistence` and this function calling `validateTxExistence` sounds weird.

contracts/common/mixin/RootChainValidator.sol
```diff
@@ -1,95 +1,78 @@
...
-  using RLP for RLP.Iterator;
-
-  // managers
-  DepositManager public depositManager;
-  WithdrawManager public withdrawManager;
-
-  // child chain contract
-  address public childChainContract;
+  
+  // using RLP for RLP.Iterator;
```
atvanguard: Remove comment.

contracts/common/mixin/RootChainValidator.sol
```diff
@@ -1,95 +1,78 @@
...
-    bytes txBytes,
-    bytes txProof
+    bytes memory txBytes,
+    bytes memory txProof
   ) public view returns (bool) {
     // fetch header block
     bytes32 headerRoot;
     uint256 start;
-    (headerRoot, start,,) = IRootChain(rootChain).headerBlock(headerNumber);
+    (headerRoot, start,,) = rootChain.getHeaderBlock(headerNumber);
```
atvanguard: Access through the auto-generated accessor for public vars.

contracts/common/mixin/RootChainValidator.sol
```diff
@@ -112,20 +95,20 @@ contract RootChainValidator is RootChainable, Lockable {
   // validate tx
   function validateTx(
```
atvanguard: if func is calling `validateTxReceiptExistence`. If this func is supposed to do both, then let's call it something else.

contracts/common/mixin/RootChainValidator.sol
```diff
@@ -151,18 +134,18 @@ contract RootChainValidator is RootChainable, Lockable {
       return false;
     }
 
-    RLP.RLPItem memory rlpItem = txBytes.toRLPItem();
+    RLPReader.RLPItem memory rlpItem = txBytes.toRlpItem();//.toList();
     if (!rlpItem.isList() || rlpItem.toList().length != 9) {
       return false;
```
atvanguard: Do we want to return false or revert?

contracts/common/mixin/RootChainValidator.sol
```diff
@@ -186,32 +169,33 @@ contract RootChainValidator is RootChainable, Lockable {
...
-    RLP.RLPItem[] memory txList = txBytes.toRLPItem().toList();
+    RLPReader.RLPItem[] memory txList = txBytes.toRlpItem().toList();
 
     // get tx sender
     return getTxSender(txList);
   }
 
   // get tx sender
-  function getTxSender(RLP.RLPItem[] txList) internal view returns (address) {
+  function getTxSender(RLPReader.RLPItem[] memory txList) internal view returns (address) {
```
atvanguard: Why keep an overload? Let's move this functionality to function above.

contracts/common/mixin/RootChainValidator.sol
```diff
@@ -186,32 +169,33 @@ contract RootChainValidator is RootChainable, Lockable {
...
     // raw tx
     bytes[] memory rawTx = new bytes[](9);
     for (uint8 i = 0; i <= 5; i++) {
-      rawTx[i] = txList[i].toData();
+      rawTx[i] = txList[i].toBytes();
     }
+
     rawTx[4] = hex"";
-    rawTx[6] = IRootChain(rootChain).networkId();
+    rawTx[6] = IWithdrawManager(registry.getWithdrawManagerAddress()).networkId(); // temp fix me
```
atvanguard: How about keeping the `networkId` var in registry?

contracts/common/mixin/RootChainValidator.sol
```diff
@@ -1,95 +1,78 @@
-pragma solidity ^0.4.24;
+pragma solidity ^0.5.2;
```
atvanguard: Looks like this file can be made into a lib.

contracts/common/mixin/RootChainable.sol
```diff
@@ -1,14 +1,13 @@
-pragma solidity ^0.4.24;
+pragma solidity ^0.5.2;
```
atvanguard: I don't think we need this file anymore.

contracts/common/tokens/StandardToken.sol
```diff
@@ -0,0 +1,38 @@
+pragma solidity ^0.5.2;
```
atvanguard: What is this file for?

contracts/common/mixin/RootChainable.sol
0xAshish: let's see for now including it here, I am kinda rethinking some wiring so will remove if not useful.

contracts/common/mixin/RootChainValidator.sol
0xAshish: yup the intention of `ChildChainVerifier` was to reuse the same here.

contracts/common/tokens/StandardToken.sol
0xAshish: test token contract for ERC223 ContractReciver.

contracts/common/mixin/RootChainValidator.sol
atvanguard: Then let's make it point to that?

contracts/common/mixin/RootChainable.sol
atvanguard: Curious to know what you have in mind.

contracts/proofs/ExitValidator.sol
```diff
@@ -66,7 +71,7 @@ contract ExitValidator is RootChainValidator {
     address owner;
     uint256 amount;
     bool burnt;
-    (owner,,amount, burnt) = withdrawManager.getExit(exitId);
+    (owner,,amount, burnt) = IWithdrawManager(registry.getWithdrawManagerAddress()).getExit(exitId);
```
atvanguard: Maybe we should pass `_withdrawManager` in constructor. That's why we made it upgradeable.

contracts/root/RootChain.sol
```diff
@@ -103,4 +130,17 @@ contract RootChain is RootChainStorage, IRootChain {
     headerBlock.createdAt = now;
     headerBlock.proposer = msg.sender;
   }
+  //   emit ChildChainChanged(childChainContract, newChildChain);
+
+  function addProofValidator(address _validator) public /* onlyOwner */  {
```
atvanguard: Let's move this part to registry if it's being accessed at more than one place.

contracts/proofs/ExitValidator.sol
0xAshish: what's correlation between passing _withdrawManager and `we made it upgradeable`?

contracts/root/RootChain.sol
0xAshish: yeah, cool did that.

contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -24,6 +24,11 @@ contract WithdrawManager is WithdrawManagerStorage /* , IWithdrawManager */ {
   using RLPReader for RLPReader.RLPItem;
   using Merkle for bytes32;
 
+  modifier isProofValidator() {
```
atvanguard: cool! The only reason I thought it should move to registry is because it is being used at multiple places.

contracts/proofs/ExitValidator.sol
atvanguard: We are accessing vars from registry
1. that can change
2. Are being used at more than 1 place.
Since WM is upgradeable, we could keep it in constructor. anyway not clear what would be better so ok for now.

contracts/common/mixin/RootChainValidator.sol
```diff
@@ -47,13 +47,16 @@ contract RootChainValidator is RootChainable, Lockable {
...
-
+    uint256 x;
+    address xx;
+    (headerRoot, start, ,,) = rootChain.headerBlocks(headerNumber);
+    return true;
     // check if tx's block is included in header and tx is in block
-    return keccak256(abi.encodePacked(blockNumber, blockTime, txRoot, receiptRoot))
-      .checkMembership(blockNumber - start, headerRoot, headerProof) &&
-    MerklePatriciaProof.verify(txBytes, path, txProof, txRoot);
+    // return keccak256(abi.encodePacked(blockNumber, blockTime, txRoot, receiptRoot))
```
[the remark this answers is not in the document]
0xAshish: 🤦‍♂️ oops.

contracts/root/withdrawManager/WithdrawManager.sol
0xAshish: em!

PR #71, opened 2019-05-01
Fix process exit

0xAshish:


contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -326,7 +326,23 @@ contract WithdrawManager is WithdrawManagerStorage /* , IWithdrawManager */ {
...
+          let ptr := mload(64)
+          // keccak256('transferAmount(address,address,uint256)') & 0xFFFFFFFF00000000000000000000000000000000000000000000000000000000
+          mstore(ptr, 0x01f4747100000000000000000000000000000000000000000000000000000000)
+          mstore(add(ptr, 4), and(_token, 0xffffffffffffffffffffffffffffffffffffffff)) // zero-out the unused bytes
+          mstore(add(ptr, 36), and(exitOwner,0xffffffffffffffffffffffffffffffffffffffff))
+          mstore(add(ptr, 68), amount) // TODO: read directly from struct
+
+          let result := call(_gas, depositManager, 0, ptr, 100, ptr, 32)
+
+          if eq(result, 0) { // if false revert
```
atvanguard: 1. Verify that `result` is 0 on revert - solidity docs don't specifically mention (they just talk about outOfGas)
2. Check both output area `(ptr, 32)` and return value `result` to be truthy.

contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -326,7 +326,23 @@ contract WithdrawManager is WithdrawManagerStorage /* , IWithdrawManager */ {
           delete ownerExits[keccak256(abi.encodePacked(_token, currentExit.owner))];
         }
 
-        IDepositManager(registry.getDepositManagerAddress()).transferAmount(_token, exitOwner, currentExit.receiptAmountOrNFTId);
+        address depositManager = registry.getDepositManagerAddress(); // TODO: make assembly call and reuse memPtr
+        uint256 amount = currentExit.receiptAmountOrNFTId;
+        uint256 _gas = gasLimit; // can't read gloablvars in asm
```
atvanguard: I think `_gas` should be able to do with less gas than `gasLimit` - (which denotes the gas quota for the entire while block)

Also, account for the gas needed for `delegateCall` in `depositManager`

contracts/root/withdrawManager/WithdrawManager.sol
atvanguard:
```
assembly {
            let ptr := mload(0x40)
            // keccak256('transferAmount(address,address,uint256)') & 0xFFFFFFFF00000000000000000000000000000000000000000000000000000000
            mstore(ptr, 0x01f4747100000000000000000000000000000000000000000000000000000000)
            mstore(add(ptr, 4), add(_token,0))
            mstore(add(ptr, 36), add(_token1,0))
            mstore(add(ptr, 68), add(amount,0))
            
            let ans := add(ptr, 100)
            let result := call(gas, depositManager, 0, ptr, 100, ans, 32)
            if eq(and(result, mload(ans)), 0) {
                revert(0,0)
            }
}
```

contracts/root/withdrawManager/WithdrawManager.sol
atvanguard: Hardcode _gas value here, I think 150k

contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -326,7 +326,23 @@ contract WithdrawManager is WithdrawManagerStorage /* , IWithdrawManager */ {
...
 
-        IDepositManager(registry.getDepositManagerAddress()).transferAmount(_token, exitOwner, currentExit.receiptAmountOrNFTId);
+        address depositManager = registry.getDepositManagerAddress(); // TODO: make assembly call and reuse memPtr
+        uint256 amount = currentExit.receiptAmountOrNFTId;
+        uint256 _gas = gasLimit.sub(52000); // sub fixed processExit cost , ::=> can't read gloablvars in asm
+        assembly {
+          let ptr := mload(64)
+          // keccak256('transferAmount(address,address,uint256)') & 0xFFFFFFFF00000000000000000000000000000000000000000000000000000000
+          mstore(ptr, 0x01f4747100000000000000000000000000000000000000000000000000000000)
+          mstore(add(ptr, 4), and(_token, 0xffffffffffffffffffffffffffffffffffffffff)) // zero-out the unused bytes
```
atvanguard: Isn't `and` with identity element redundant?

contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -326,7 +326,23 @@ contract WithdrawManager is WithdrawManagerStorage /* , IWithdrawManager */ {
...
+          mstore(add(ptr, 36), and(exitOwner,0xffffffffffffffffffffffffffffffffffffffff))
+          mstore(add(ptr, 68), amount) // TODO: read directly from struct
+          let ret := add(ptr,100)
+          let result := call(_gas, depositManager, 0, ptr, 100, ret, 32) // returns 1 if success
+
+          if eq(and(result,mload(ret)), 0) { // if false=>  revert & return bool
```
atvanguard: fix comment

PR #72, opened 2019-05-06
remove transferFrom, emit LogTransfer in transferWithSig

atvanguard:
Also strips away the need to approve token transfers


contracts/child/misc/ParentTokenMock.sol
```diff
@@ -1,19 +0,0 @@
-pragma solidity ^0.5.2;
-
-import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
-
-import "./IParentToken.sol";
-// demo token parent contract  
-
-
-contract ParentTokenMock is IParentToken, Ownable {
```
0xAshish: please add it back, it's used to test `IParentToken` interface and also a sample implementation of it.
feel free to rename it.

atvanguard: Added back.

Issue #48, opened 2019-01-17
Add a way to exit with in-flight tx


jdkanani:

jdkanani:
@0xAshish I remember you were explaining a way to replace exit. Let's discuss.

jdkanani:
Here is my first attempt.

1. `A` is transferring tokens to `B` in `tx1`
2. `tx1` is in-flight or limbo tx
3. `A` or `B` or user who knows `tx1`, starts a special exit by submitting `tx1` on mainchain contract by referencing `tx0` which was before `tx1` - and placing `exit bond`. `tx0` will have `A`'s updated balance. This starts exit period 1.
5. In period 1, `B` has to claim the balance by making simple transaction on root chain contract and it will start exit for `B` with (priority + 1).
4. In period 2, anyone watching the chain (mostly current users & validators) can challenge if `A` or `B` made tx after that using nonce and root hash. If challenge is successful, corresponding user will not be able to exit.

PR #73, opened 2019-05-14
MoreVP - Exit with burn in-flight

atvanguard:
Exit from a burn in-flight tx by referencing an older tx.
https://github.com/maticnetwork/contracts/issues/48


contracts/root/lib/ExitTxValidator.sol
```diff
@@ -0,0 +1,65 @@
...
+    require(
+      childToken == RLPReader.toAddress(txList[3]), // corresponds to "to" field in tx)
+      "Reference and exit tx do not correspond to the same token"
+    );
+    bytes memory txData = RLPReader.toBytes(txList[5]);
+    bytes4 funcSig = BytesLib.toBytes4(BytesLib.slice(txData, 0, 4));
+    if (funcSig == WITHDRAW_FUNC_SIG) {
+      require(txData.length == 36, "Invalid tx"); // 4 bytes for funcSig and a single bytes32 parameter
+      exitAmountOrTokenId = BytesLib.toUint(txData, 4);
+      burnt = true;
```
jdkanani: Make sure we don't set burnt true for every transaction.

contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -31,6 +32,73 @@ contract WithdrawManager is WithdrawManagerStorage /* , IWithdrawManager */ {
...
+      address(registry)
+    );
+    require(
+      registry.rootToChildToken(rootToken) == childToken,
+      "INVALID_ROOT_TO_CHILD_TOKEN_MAPPING"
+    );
+
+    // validate exitTx
+    (uint256 exitAmountOrTokenId, bool burnt) = ExitTxValidator.processExitTx(exitTx, childToken, msg.sender);
+    // gives stack too deep, fix later
```
jdkanani: Fix stack too deep :)

contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -237,8 +305,8 @@ contract WithdrawManager is WithdrawManagerStorage /* , IWithdrawManager */ {
     uint256 _exitId = (
       headerNumber * HEADER_BLOCK_NUMBER_WEIGHT +
       withdrawBlockNumber * WITHDRAW_BLOCK_NUMBER_WEIGHT +
-      path.toRlpItem().toBytes().toRlpItem().toUint() * 100000 +
-      oIndex
+      path.toRlpItem().toBytes().toRlpItem().toUint() * 100000
+      // + oIndex
```
jdkanani: Will exit Ids be the same?

contracts/root/withdrawManager/WithdrawManager.sol
atvanguard: Have fixed this along with some changes in the next PR.

contracts/root/lib/ExitTxValidator.sol
atvanguard: For sure :)

contracts/root/withdrawManager/WithdrawManager.sol
atvanguard: Yes, the way we calculate the *age* will remain same like in MVL. Just that we will take the age of reference tx now. I'll probably give an exit ID that'll be different from *age*, like sequential maybe - Yet to think about the *implementation details* of the challenges, so using age right now.

PR #88, opened 2019-07-06
Support bulk deposits

atvanguard:
Gas requirements, originally
```
depositErc20 - 165227
depositErc721 - 182304
```

After refactoring (moving `mapping(uint256 => DepositBlock) public deposits` to `DepositManager`)
```
depositErc20 - 143890
depositErc721 - 160989
(143890 + 160989) * 15 = 4.57M

Bulk deposit gas for 15 erc20 and erc721 transfers = 3.68M
```
20% savings in gas and improved UX! 🍾


0xAshish:
we should implement 1155 right away for bulk deposits.

contracts/root/depositManager/IDepositManager.sol
```diff
@@ -5,4 +5,6 @@ interface IDepositManager {
   function transferAssets(address _token, address  _user, uint256 _amountOrNFTId) external;
   function depositERC20(address _token, uint256 _amount) external;
   function depositERC721(address _token, uint256 _tokenId) external;
+
+  function onERC20Received(address _user, address _token, uint256 _amount) external;
```
0xAshish: not really an issue but `onERC20Received ` should come from any interface. 😄

contracts/root/depositManager/DepositHelper.sol
```diff
@@ -0,0 +1,52 @@
...
+
+import { Registry } from "../../common/Registry.sol";
+import { IDepositManager } from "./IDepositManager.sol";
+
+contract DepositHelper {
+
+  Registry private registry;
+
+  enum Type { Invalid, ERC20, ERC721 }
+  mapping(address => Type) public tokenType;
```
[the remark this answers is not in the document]
atvanguard: Yeah, but I wanted to avoid calling an external contract for every transfer! But I guess there is no other way because even this cache will be outdated if we un-map a token.

contracts/root/depositManager/DepositManager.sol
```diff
@@ -26,6 +26,14 @@ contract DepositManager is DepositManagerStorage, IDepositManager, IERC721Receiv
     _;
   }
 
+  modifier onlyDepositHelper() {
```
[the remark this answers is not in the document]
atvanguard: How will it be a library?
We are calling `transferBatch` in the depositHelper. Bdw I am thinking of including these functions directly in the depositManager instead of having another contract

contracts/root/depositManager/IDepositManager.sol
atvanguard: ?

contracts/common/Registry.sol
```diff
@@ -118,6 +118,11 @@ contract Registry is Ownable {
     return rootToChildToken[_token] != address(0x0);
   }
 
+  function isTokenMappedAndIsErc721(address _token) public view returns (bool) {
+    require(isTokenMapped(_token), "TOKEN_NOT_MAPPED");
```
0xAshish: please remove require condition.

contracts/root/depositManager/DepositManager.sol
```diff
@@ -49,6 +49,23 @@ contract DepositManager is DepositManagerStorage, IDepositManager, IERC721Receiv
...
+    require(
+      _tokens.length == _amountOrTokens.length,
+      "Invalid Input"
+    );
+    for (uint256 i = 0; i < _tokens.length; i++) {
+      // will revert if token is not mapped
+      if (registry.isTokenMappedAndIsErc721(_tokens[i])) {
+        depositERC721ForUser(_tokens[i], _user, _amountOrTokens[i]);
+      } else {
+        depositERC20ForUser(_tokens[i], _user, _amountOrTokens[i]);
```
0xAshish: will/should never come here if require in fails in `isTokenMappedAndIsErc721`(will revert).

atvanguard: Will not come here if reverts, but it will reach here if isTokenMappedAndIsErc721 returns false.

contracts/common/Registry.sol
atvanguard: No, I want it to revert. Otherwise it will go to else block even if the token is not mapped.
```
if (registry.isTokenMappedAndIsErc721(_tokens[i])) {
        depositERC721ForUser(_tokens[i], _user, _amountOrTokens[i]);
      } else {
        depositERC20ForUser(_tokens[i], _user, _amountOrTokens[i]);
}
```

0xAshish: aha! cool

contracts/root/depositManager/DepositManager.sol
```diff
@@ -79,47 +96,52 @@ contract DepositManager is DepositManagerStorage, IDepositManager, IERC721Receiv
...
-  function _createDepositBlock(address _user, address _token, uint256 amountOrNFTId)
+  function _safeCreateDepositBlock(address _user, address _token, uint256 _amountOrToken)
     internal
     isTokenMapped(_token)
   {
-    rootChain.createDepositBlock(_user, _token, amountOrNFTId);
+    rootChain.createDepositBlock(_token, _amountOrToken, _user);
+  }
+
+  function _createDepositBlockBulk(address[] memory _tokens, uint256[] memory _amountOrTokens, address _user)
```
0xAshish: where are we using it `_createDepositBlockBulk` ?

atvanguard: to call `rootChain.bulkCreateDepositBlocks`

atvanguard: Whoops missed it. Check it out now. I removed this function.

PR #90, opened 2019-07-06
Bonded exits

### LINKED ISSUE #17: Add bond for exit and reward challenger
While starting an exit, an exitNFT's owner will put small bond and challenger will get that bond if challenge gets accepted.

Question - is there any way we can avoid exit bond?

atvanguard:
Closes #17

### DIFF
--- a/contracts/root/withdrawManager/WithdrawManagerStorage.sol
+++ b/contracts/root/withdrawManager/WithdrawManagerStorage.sol
@@ -11,13 +11,12 @@ contract ExitsDataStructure {
   }
 
   struct PlasmaExit {
-    address owner;
-    address token;
     uint256 receiptAmountOrNFTId;
     bytes32 txHash;
-    bool burnt;
+    address owner;
+    address token;
     address predicate;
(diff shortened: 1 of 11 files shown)


contracts/root/withdrawManager/WithdrawManagerStorage.sol
```diff
@@ -17,6 +17,7 @@ contract ExitsDataStructure {
     bytes32 txHash;
     bool burnt;
     address predicate;
+    uint256 inputCount;
```
0xAshish: @atvanguard if possible please restructure this `struct` 
```
struct PlasmaExit {
    uint256 receiptAmountOrNFTId;
    uint256 inputCount;
    bytes32 txHash;
    bytes32 txHash;
    address owner;
    address token;
    address predicate;
    address predicate;
    bool burnt;
    bool burnt;
    // Mapping from age of input to Input
    mapping(uint256 => Input) inputs;
  }
```
possibly it will pack our vars into multiple 32 bytes slots.

PR #93, opened 2019-07-12
Start exit directly from deposited tokens

atvanguard:


contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -105,6 +110,23 @@ contract WithdrawManager is WithdrawManagerStorage, IWithdrawManager {
     _;
   }
 
+  function startExitWithDepositedTokens(uint256 depositId, address token, uint256 amountOrToken)
+    external
+    payable
+    isBondProvided
```
0xAshish: why `isBondProvided `?
we need to make sure that there is at least 1 checkpoint before `ExitWithDepositedTokens`.

contracts/common/Registry.sol
```diff
@@ -123,6 +136,13 @@ contract Registry is Ownable {
     return isERC721[_token];
   }
 
+  function isTokenMappedAndGetPredicate(address _token) public view returns (address) {
+    if (isTokenMappedAndIsErc721(_token)) {
+      return erc721Predicate;
```
0xAshish: somehow I don't like this method. let's say if some erc20/erc721 has multiple predicates?

contracts/root/withdrawManager/WithdrawManager.sol
atvanguard: This needs to be a bonded exit, since I could have spent the tokens on the side chain and start an exit here; then that needs to be challenged.

contracts/common/Registry.sol
atvanguard: Then we will return the current 20/721 predicate specifically. This predicate needs to be mentioned in the exit to process challenges, if any.

contracts/root/withdrawManager/WithdrawManager.sol
0xAshish: Waiting on 1 checkpoint still sounds better than, bonded exit.
let's say user accidentally deposited wrong amount(more then needed) and he just wants the normal exit to get this amount out on top of that now also need to transfer bond amount, which will be locked for some time(in case of `ExitWithDepositedTokens` we should think of faster exit then normal plasma exit if possible).

0xAshish: @jdkanani what do you think?.

PR #97, opened 2019-07-25
update deploy_migrations and README

atvanguard:


contracts/common/Registry.sol
```diff
@@ -61,10 +61,6 @@ contract Registry is Ownable {
       _rootToken != address(0x0) && _childToken != address(0x0),
       "INVALID_TOKEN_ADDRESS"
     );
-    require(
```
0xAshish: ?

0xAshish:
please don't merge it.

contracts/common/Registry.sol
atvanguard: Removed it because it is ACLed on `onlyOwner` and I would like to have the ability to remap a token 😃

PR #103, opened 2019-08-14
check transferWithSig expiration in marketplace exits/decrecation

atvanguard:


atvanguard:
> we need to think about if someone/operator tries to do fraud in between checkpoints
Please elaborate on the nature of the fraud? @0xAshish

0xAshish:
mostly operator can do that kinda fraud like doing reorg and just spending your amount using signed tx, because once you realised that given block number you are safe(user thinks) to spend amount but now operator can just execute the order and challenge you. obviously, this is for short period like the gap between checkpoint. @atvanguard

PR #118, opened 2019-09-26
Change withdraw duration to 1 minute and use @truffle/hdwallet-provider

atvanguard:


contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -384,6 +384,7 @@ contract WithdrawManager is WithdrawManagerStorage, IWithdrawManager {
   }
 
   function getExitableAt(uint256 createdAt) internal view returns (uint256) {
-    return Math.max(createdAt + 2 * HALF_EXIT_PERIOD, now + HALF_EXIT_PERIOD);
```
0xAshish: just change the constant Var `HALF_EXIT_PERIOD`.

package.json
```diff
@@ -74,7 +75,6 @@
     "ganache-cli": "6.5.0",
     "merkle-patricia-tree": "2.3.2",
     "openzeppelin-solidity": "2.2.0",
-    "solidity-rlp": "^2.0.0",
-    "truffle-hdwallet-provider": "^1.0.17"
```
0xAshish: why remove @?
https://stackoverflow.com/questions/36667258/what-is-the-meaning-of-the-at-prefix-on-npm-packages?noredirect=1&lq=1

contracts/root/withdrawManager/WithdrawManager.sol
atvanguard: Then it can't be used as a replacement for a currently deployed contract. We can't change storage.

package.json
atvanguard: What? I added `@truffle/hdwallet-provider` and removed `truffle-hdwallet-provider` which is deprecated.

PR #119, opened 2019-09-27
replace user by signer staked event (#117)

atvanguard:
* replace user by signer staked event

* fix tests


atvanguard:
CI: https://travis-ci.org/maticnetwork/contracts/builds/590361930

PR #123, opened 2019-10-09
Rewards distribution

0xAshish:
- rewards distribution accounting is done on Heimdall  layer
- with each checkpoint updated Merkle root of accounts is submitted
- in order to claim rewards, validators submits proof of accounting

### DIFF
--- a/contracts/staking/StakeManager.sol
+++ b/contracts/staking/StakeManager.sol
@@ -8,7 +8,7 @@ import { RLPReader } from "solidity-rlp/contracts/RLPReader.sol";
 
 import { BytesLib } from "../common/lib/BytesLib.sol";
 import { ECVerify } from "../common/lib/ECVerify.sol";
-import { MerklePatriciaProof } from "../common/lib/MerklePatriciaProof.sol";
+import { Merkle } from "../common/lib/Merkle.sol";
 import { Lockable } from "../common/mixin/Lockable.sol";
 import { RootChainable } from "../common/mixin/RootChainable.sol";
 import { Registry } from "../common/Registry.sol";
@@ -20,6 +20,7 @@ import { ValidatorContract } from "./Validator.sol";
 contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
   using SafeMath for uint256;
   using ECVerify for bytes32;
+  using Merkle for bytes32;
   using RLPReader for bytes;
   using RLPReader for RLPReader.RLPItem;
 
@@ -175,17 +176,17 @@ contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
     emit ReStaked(validatorId, validators[validatorId].amount, totalStaked);
   }
 
-  function claimRewards(uint256 validatorId, bytes memory accountState, bytes memory path, bytes memory stateProof) public /*onlyStaker(validatorId) */ {
-    require(MerklePatriciaProof.verify(accountState, path, stateProof, accountStateRoot));
-    RLPReader.RLPItem[] memory dataList = accountState.toRlpItem().toList();
-    uint256 amount = dataList[0].toUint();
-    uint256 _reward = amount.sub(validators[validatorId].totalReward);
+  function claimRewards(uint256 validatorId, uint256 accountBalance, bytes memory path, bytes memory proof) public /*onlyStaker(validatorId) */ {
+    // accountState = keccak256(abi.encodePacked(validatorId, accountBalance))
+    require(keccak256(abi.encodePacked(validatorId, accountBalance)).checkMembership(path, accountStateRoot, proof));
+    uint256 _reward = accountBalance.sub(validators[validatorId].totalReward);
     address _contract = validators[validatorId].contractAddress;
     if (_contract == address(0x0)) {
       validators[validatorId].reward = validators[validatorId].reward.add(_reward);
     }
     else {
       // TODO: delegator bond/share rate if return needs to be updated periodically
+      // otherwise validator can delay and get all the delegators reward
       ValidatorContract(_contract).updateRewards(_reward, currentEpoch, validators[validatorId].amount);
     }
     totalRewardsLiquidated += validators[validatorId].reward;
--- a/contracts/staking/StakeManager.sol
+++ b/contracts/staking/StakeManager.sol
@@ -73,7 +73,7 @@ contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
 
   constructor (address _registry, address _rootchain) ERC721Full("Matic Validator", "MV") public {
     registry = _registry;
-    changeRootChain(_rootchain);
+    rootChain = _rootchain;
   }
 
   modifier onlyStaker(uint256 validatorId) {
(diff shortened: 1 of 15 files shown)


contracts/staking/StakeManager.sol
```diff
@@ -165,6 +175,31 @@ contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
...
+    address _contract = validators[validatorId].contractAddress;
+    if (_contract == address(0x0)) {
+      validators[validatorId].reward = validators[validatorId].reward.add(_reward);
+    }
+    else {
+      // TODO: delegator bond/share rate if return needs to be updated periodically
+      // otherwise validator can delay and get all the delegators reward
+      ValidatorContract(_contract).updateRewards(_reward, currentEpoch, validators[validatorId].amount);
+    }
+    totalRewardsLiquidated += validators[validatorId].reward;
```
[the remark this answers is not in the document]
0xAshish: yeah yeah, I did make this change before reading this comment.
Nice catch  @mankenavenkatesh Thanks

contracts/staking/StakeManager.sol
```diff
@@ -165,6 +175,31 @@ contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
     emit ReStaked(validatorId, validators[validatorId].amount, totalStaked);
   }
 
+  function claimRewards(uint256 validatorId, uint256 accountBalance, uint256 index, bytes memory proof) public /*onlyStaker(validatorId) */ {
+    // accountState = keccak256(abi.encodePacked(validatorId, accountBalance))
+    require(keccak256(abi.encodePacked(validatorId, accountBalance)).checkMembership(index, accountStateRoot, proof));
```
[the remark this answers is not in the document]
0xAshish: I had this in mind but the assumption is that it's transactors money.

atvanguard:
Has `challengeStateRootUpdate` been discussed? Going back to _challengeable_ checkpoints will change the system assumptions significantly. @mankenavenkatesh @jdkanani

0xAshish:
> Has `challengeStateRootUpdate` been discussed? Going back to _challengeable_ checkpoints might change the system assumptions significantly. @mankenavenkatesh @jdkanani

Yup, so the assumption is that after `challengeStateRootUpdate` will stop the checkpoints and only once the account state is updated with more threshold like 80-90% or even more 100% votes then only will restart the checkpoints.
To be clear this worst of worst-case where 2/3+1 colluded, we need to slash all 2/3+1 once the `challengeStateRootUpdate` has been succeeded.
also, the challenge period is a single checkpoint. since this is validators own rewards at stake they should verify it at Heimdall and mainnet layer.
Still figuring out the inner details any inputs would be appreciated 😄

atvanguard:
> The challenge period is a single checkpoint.
This is the culprit. I think this period is far too low.
Attack scenario: Other validators notice a validator is down, they make his account balance 0, then expectation is validator has to come back online within a single checkpoint.

If you think about about having longer challenge durations, what happens to checkpoints made after the malicious one needs to be thought about and how will plasma exits be affected (if at all).

vaibhavchellani:
>Attack scenario: Other validators notice a validator is down, they make his account balance 0, then expectation is validator has to come back online within a single checkpoint.

The assumption here is that 2/3+1 validators collude. It should be quite possible that people can challenge anytime. Also this model relies on 2/3+1 security so this attack is possible for sure and there will be fund loss. Having a challenging period of 1 checkpoint is something we don't `have` to do but is just a fallback for very small validators.

0xAshish:
@atvanguard please push this PR by today, thanks 🌚

contracts/staking/StakeManager.sol
```diff
@@ -165,6 +176,40 @@ contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
...
+    if (_contract == address(0x0)) {
+      validators[validatorId].reward = validators[validatorId].reward.add(_reward);
+    }
+    else {
+      // TODO: delegator bond/share rate if return needs to be updated periodically
+      // otherwise validator can delay and get all the delegators reward
+      ValidatorContract(_contract).updateRewards(_reward, currentEpoch, validators[validatorId].amount);
+    }
+    totalRewardsLiquidated += _reward;
+    require(totalRewardsLiquidated <= totalRewards, "Oops this shouldn't have happened");// pos 2/3+1 is colluded
```
atvanguard: Error message is inconclusive.

contracts/staking/StakeManager.sol
```diff
@@ -165,6 +176,40 @@ contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
...
+    uint256 amount = validators[validatorId].reward;
+    address _contract = validators[validatorId].contractAddress;
+    if (_contract != address(0x0)) {
+      amount = amount.add(ValidatorContract(_contract).withdrawRewardsValidator());
+    }
+    validators[validatorId].reward = 0;
+    require(token.transfer(msg.sender, amount), "Insufficent rewards");
+  }
+
+  function delegationTransfer(uint256 amount, address delegator) public {
```
atvanguard: Where is this being used?

contracts/staking/StakeManager.sol
0xAshish: hoping this never comes, should I add `pos 2/3+1 has colluded` to the require msg?

contracts/staking/StakeManager.sol
```diff
@@ -28,17 +31,24 @@ contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
   uint256 public EPOCH_LENGTH = 256; // unit : block
   uint256 public UNSTAKE_DELAY = DYNASTY.mul(2); // unit: epoch
 
+  // TODO: add events and gov. based update function
+  uint256 public proposerToSignerRewards = 10; // will be used with fraud proof
```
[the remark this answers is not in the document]
0xAshish: Heimdall is picking this ratio from the contract and later on, it'll also be used in `challangeStateRootUpdate`.

contracts/staking/StakeManager.sol
```diff
@@ -28,17 +31,24 @@ contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
...
+  uint256 public totalRewards;
+  uint256 public totalRewardsLiquidated;
+  bytes32 public accountStateRoot;
 
   enum Status { Inactive, Active, Locked }
 
   struct Validator {
     uint256 amount;
     uint256 reward;
+    uint256 totalReward;
```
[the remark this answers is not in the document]
0xAshish: done

contracts/staking/StakeManager.sol
0xAshish: ah! forgot to use it in delegation manager.

0xAshish: Thanks

contracts/staking/StakeManager.sol
```diff
@@ -165,6 +176,40 @@ contract StakeManager is Validator, IStakeManager, RootChainable, Lockable {
...
+    if (_contract == address(0x0)) {
+      validators[validatorId].reward = validators[validatorId].reward.add(_reward);
+    }
+    else {
+      // TODO: delegator bond/share rate if return needs to be updated periodically
+      // otherwise validator can delay and get all the delegators reward
+      ValidatorContract(_contract).updateRewards(_reward, currentEpoch, validators[validatorId].amount);
+    }
+    totalRewardsLiquidated += _reward;
+    require(totalRewardsLiquidated <= totalRewards, "Liquidating more rewards then checkpoints submitted");// pos 2/3+1 is colluded
```
atvanguard: Lol, it's still incorrect. Nevermind.

contracts/staking/StakeManager.sol
atvanguard: Lol, it's still incorrect. Nevermind.

Issue #115, opened 2019-09-23
MAT-327 Setup geth instead of parity to run the child chain on locally


atvanguard:

jdkanani:
Any reason for that?

atvanguard:
Yes, to be sure our tests pass for the geth running on our custom config + now we have made changes to bor so tests will need that e.g. `LogFeeTransfer` event.

jdkanani:
PR #127

PR #126, opened 2019-10-21
MAT-250: Matic child token / Fee updates

atvanguard:


atvanguard:
Will add tests once #115 is resolved.

contracts/root/predicates/ERC20Predicate.sol
```diff
@@ -433,8 +441,8 @@ contract ERC20Predicate is IErcPredicate {
         "Withdrawer and referenced tx do not match"
       );
       closingBalance = BytesLib.toUint(logData, 64); // output1
-    } else if (eventSignature == LOG_TRANSFER_EVENT_SIG) {
-      // event LogTransfer(
+    } else if (eventSignature == LOG_TRANSFER_EVENT_SIG || eventSignature == LOG_FEE_TRANSFER_EVENT_SIG) {
```
0xAshish: `LOG_FEE_TRANSFER_EVENT_SIG` sig is present in each and every transaction !?

atvanguard: After the Bor changes, yes.

test/root/predicates/predicateTestUtils.js
```diff
@@ -32,13 +32,15 @@ export function getAge(utxo) {
...
-export async function getRawInflightTx(fn, from, web3, gas) {
-  const options = { from, gas: gas || 30000 } // minimal: 26136 is required, not sure why
+export async function getRawInflightTx(fn, from, web3, gas, options = {}) {
+  Object.assign(options, { from: from || options.from, gas: gas || 30232 }) // minimal: 30232 is required, not sure why
   try {
     await fn(options)
     assert.fail('should have failed')
   } catch (e) {
-    const txHash = /transactionHash": "(0[xX][0-9a-fA-F]+)"/g.exec(e.hijackedStack)[1]
+    // console.log(e)
```
0xAshish: remove dead code.

test/root/predicates/predicateTestUtils.js
```diff
@@ -32,13 +32,15 @@ export function getAge(utxo) {
 
 // this is hack to generate a raw tx that is expected to be inflight
```
0xAshish: you can generate rawTX using web3 why go with this hack?

contracts/root/predicates/ERC20Predicate.sol
```diff
@@ -504,11 +525,11 @@ contract ERC20Predicate is IErcPredicate {
   {
     bytes4 funcSig = BytesLib.toBytes4(BytesLib.slice(txData, 0, 4));
     if (funcSig == WITHDRAW_FUNC_SIG) {
-      require(txData.length == 36, "Invalid tx"); // 4 bytes for funcSig and a single bytes32 parameter
+      // require(txData.length == 36, "Invalid tx"); // 4 bytes for funcSig and a single bytes32 parameter
```
0xAshish: why ?

contracts/child/MaticChildERC20.sol
```diff
@@ -75,22 +77,19 @@ contract MaticChildERC20 is BaseERC20 {
     return account.balance;
   }
 
+  function _transfer(address sender, address recipient, uint256 amount) internal {
```
0xAshish: why do we need to override _transfer function ?

contracts/root/predicates/ERC20Predicate.sol
atvanguard: This was interesting. So, until now we were assuming say a withdraw tx on matic to be 36 bytes (4 bytes for funcSig and a single bytes32 parameter). 
Now, a `withdraw` in MaticChildErc20 also involves sending ether (msg.value) and tx payload size becomes bigger. So, for this predicate to work for maticEth, I removed this check because it was inconclusive anyway. Do you see an attack vector?

test/root/predicates/predicateTestUtils.js
atvanguard: Yes, I got a little lazy while writing this one (since it involves instantiating a contract again, passing nonce etc and doesn't let me use truffle artifacts.) So thought of this quick hack 😅

test/root/predicates/predicateTestUtils.js
atvanguard: Done.

0xAshish:
@atvanguard why `MaticToken` isn't erc20 anymore? 
`BaseERC20` doesn't make any sense now, since we have removed `ERC20` from `MaticToken` we should put all the code in a single contract file in `MaticToken` contract and revert the `BaseERC20`.

contracts/root/predicates/ERC20Predicate.sol
```diff
@@ -504,11 +525,9 @@ contract ERC20Predicate is IErcPredicate {
   {
     bytes4 funcSig = BytesLib.toBytes4(BytesLib.slice(txData, 0, 4));
     if (funcSig == WITHDRAW_FUNC_SIG) {
-      require(txData.length == 36, "Invalid tx"); // 4 bytes for funcSig and a single bytes32 parameter
       amount = BytesLib.toUint(txData, 4);
       exitType = ExitType.Burnt;
     } else if (funcSig == TRANSFER_FUNC_SIG) {
```
0xAshish: `require(txData.length == 36, "Invalid tx");` these checks were for collision attacks with function sig(due to smaller size 4bytes), will need to be always carefull with new contracts additions.

contracts/root/predicates/ERC20Predicate.sol
```diff
@@ -353,9 +365,16 @@ contract ERC20Predicate is IErcPredicate {
     bytes memory logData = inputItems[2].toBytes();
     inputItems = inputItems[1].toList(); // topics
     // now, inputItems[i] refers to i-th (0-based) topic in the topics array
-    // inputItems[0] is the event signature
+    bytes32 eventSignature = bytes32(inputItems[0].toUint()); // inputItems[0] is the event signature
     data.rootToken = address(RLPReader.toUint(inputItems[1]));
-    // rootToken = RLPReader.toAddress(inputItems[1]); // investigate why this reverts
+
+    // When child chain is started, since child matic is a genenis contract at 0x1010,
+    // the root token corresponding to matic is not known, hence child token address is emitted in LogFeeTransfer events.
```
0xAshish: what!!?? no,  Matic chain shouldn't start without `Matic` tokens mapping.

test/root/predicates/ERC20Predicate.test.js
```diff
@@ -88,7 +118,7 @@ contract('ERC20Predicate', async function(accounts) {
         contracts.ERC20Predicate.startExitForOutgoingErc20Transfer, headerNumber, blockProof, block.number, block.timestamp, reference, 1, /* logIndex */ exitTx)
       const logs = logDecoder.decodeLogs(startExitTx.receipt.rawLogs)
       // console.log(startExitTx, logs)
-      const log = logs[1]
+      const log = logs[utils.filterEvent(logs, 'ExitStarted')]
```
0xAshish: lol this next line is stupid 😉 `logs[utils.filterEvent(logs, 'ExitStarted')]==log.event.should.equal('ExitStarted')`.

test/root/predicates/ERC20Predicate.test.js
```diff
@@ -116,7 +146,7 @@ contract('ERC20Predicate', async function(accounts) {
       const startExitTx = await utils.startExitForErc20PredicateLegacy(contracts.ERC20Predicate.startExitForOutgoingErc20Transfer, headerNumber, blockProof, block.number, block.timestamp, reference, 1, /* logIndex */ exitTx)
       const logs = logDecoder.decodeLogs(startExitTx.receipt.rawLogs)
       // console.log(startExitTx, logs)
-      const log = logs[1]
+      const log = logs[utils.filterEvent(logs, 'ExitStarted')]
```
0xAshish: this should be hardcoded, we should expect things at the exact location. since it's a test it doesn't matter regarding those hardcoded values.

contracts/root/predicates/ERC20Predicate.sol
atvanguard: Assertions on length aren't doing us much good.

0xAshish: yeah, although we should be careful with new token mappings.

PR #127, opened 2019-10-21
MAT-327 Use geth instead of parity

jdkanani:


0xAshish:
@atvanguard @jdkanani i'll try to fix the CI tonight. I know why it's failing.

PR #128, opened 2019-10-22
MAT-347: Fix Geth config for tests to pass

atvanguard:


atvanguard:
I get
```
WARN [10-22|12:48:00.176] Block sealing failed                     err="unauthorized signer"
```
and txs on geth are not getting mined due to something I changed, I think.

PR #136, opened 2019-11-11
MAT-400: Bring back onFinalizeExit + MintableERC721Predicate

atvanguard:


contracts/root/predicates/ERC721Predicate.sol
```diff
@@ -199,12 +176,13 @@ contract ERC721Predicate is IErcPredicate {
     return ageOfChallengeTx > age;
   }
 
-  // function onFinalizeExit(address exitor, address token, uint256 tokenId)
-  //   external
-  //   onlyWithdrawManager
-  // {
-  //   depositManager.transferAssets(token, exitor, tokenId);
-  // }
+  function onFinalizeExit(bytes calldata data)
```
0xAshish: this method is duplicated 3-4 times why not keep it in `Ipredicate` or `MixinPredicate`

test/helpers/artifacts.js
```diff
@@ -28,7 +29,7 @@ export const TestToken = artifacts.require('TestToken')
 export const DummyERC20 = artifacts.require('DummyERC20')
 export const RootERC721 = artifacts.require('RootERC721')
 export const ERC721PlasmaMintable = artifacts.require('ERC721PlasmaMintable')
-export const ExitNFT = artifacts.require('ExitNFT')
+export const ExitNFT = artifacts.require('ExitNFT.sol')
```
0xAshish: why `.sol` extension is needed ?

contracts/root/predicates/ERC721Predicate.sol
atvanguard: Moved to `PredicateUtils` - since other predicates already inherit from it.

PR #137, opened 2019-11-13
Artifacts for beta-mainnet and helper libs

atvanguard:


atvanguard:
No need for files in `contracts-core/artifacts/` to be committed to the repo. Best to copy these over before releasing; just like building a typescript project before publishing to npm.

PR #143, opened 2019-11-25
MAT-462: Pass Heimdall Id to rootChain constructor

atvanguard:


migrations/2_deploy_root_contracts.js
```diff
@@ -144,7 +144,7 @@ module.exports = async function(deployer, network) {
     console.log('deploying contracts...')
     await deployer.deploy(Registry)
     await Promise.all([
-      deployer.deploy(RootChain, Registry.address),
+      deployer.deploy(RootChain, Registry.address, 'heimdall-P5rXwg'),
```
vaibhavchellani: this should be `process.env.HEIMDALL_ID` ?

scripts/stake.js
```diff
@@ -20,8 +20,9 @@ async function stake(address, amount) {
 }
 
 module.exports = async function(callback) {
-  const stakeFor = '0xE0938d9fd679bB6B83bf31fA62c433646B9F749e'
-  const amount = web3.utils.toWei('50')
+  const stakeFor = process.argv[6]
+  const amount = web3.utils.toWei(process.argv[7])
+  console.log(`Staking ${amount} for ${stakeFor}...`)
   await stake(stakeFor, amount)
```
vaibhavchellani: Can we also pring `txHash` ?

atvanguard: Done.

PR #164, opened 2020-02-04
Audit fixes

atvanguard:


contracts/common/lib/BytesLib.sol
```diff
@@ -134,11 +135,13 @@ library BytesLib {
...
   // Pad a bytes array to 32 bytes
   function leftPad(bytes memory _bytes) internal pure returns (bytes memory) {
-    bytes memory newBytes = new bytes(32 - _bytes.length);
+    // may underflow if bytes.length < 32. Hence using SafeMath.sub
+    bytes memory newBytes = new bytes(SafeMath.sub(32, _bytes.length));
     return concat(newBytes, _bytes);
   }
 
   function toBytes32(bytes memory b) internal pure returns (bytes32) {
+    require(b.length >= 32, "Bytes array should atleast be 32 bytes");
```
0xAshish: we should have one function without this check because there might be a case where we might need to convert the smaller data type to bytes32.

contracts/common/lib/BytesLib.sol
```diff
@@ -155,17 +158,14 @@ library BytesLib {
   function fromBytes32(bytes32 x) internal pure returns (bytes memory) {
     bytes memory b = new bytes(32);
     for (uint i = 0; i < 32; i++) {
-      b[i] = byte(uint8(uint(x) / (2**(8*(19 - i)))));
+      b[i] = byte(uint8(uint(x) / (2**(8*(31 - i)))));
```
0xAshish: just for my information why 31?

contracts/common/lib/Common.sol
```diff
@@ -24,11 +24,8 @@ library Common {
 
   // convert uint256 to bytes
   function toBytes(uint256 _num) public pure returns (bytes memory _ret) {
-    assembly {
-      _ret := mload(0x10)
-      mstore(_ret, 0x20)
-      mstore(add(_ret, 0x20), _num)
-    }
+    _ret = new bytes(32);
```
0xAshish: This function is a duplicate of `BytesLib.fromUint`.

contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -128,14 +128,19 @@ contract WithdrawManager is WithdrawManagerStorage, IWithdrawManager {
       referenceTxData[offset + 1].toBytes() // blockProof
     );
 
+    uint256 _branchMask = branchMask.toRlpItem().toUint();
+    require(
+      _branchMask & 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF00000000 == 0,
```
0xAshish: please explain `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF00000000`, and why trailing zeros? shouldn't be all `FF`'s

contracts/common/lib/Merkle.sol
```diff
@@ -10,10 +10,11 @@ library Merkle {
   ) public pure returns (bool) {
     bytes32 proofElement;
     bytes32 computedHash = leaf;
-    uint256 len = (proof.length / 32) * 32;
+    require(proof.length % 32 == 0, "Invalid proof length");
+    // uint256 len = (proof.length / 32) * 32;
```
0xAshish: Thanks for keeping the comment explained stuff, I think now we can remove :d

contracts/common/lib/BytesLib.sol
atvanguard: Can we add that function when we actually require it?

contracts/common/lib/BytesLib.sol
atvanguard: Discussed offline.

contracts/common/lib/Common.sol
atvanguard: Removed this one.

contracts/common/lib/Merkle.sol
atvanguard: Done.

contracts/common/lib/BytesLib.sol
0xAshish: yeah sure works for me.

PR #174, opened 2020-02-10
Add housekeeping functions

atvanguard:


contracts/root/RootChain.sol
```diff
@@ -131,4 +125,13 @@ contract RootChain is RootChainStorage, IRootChain {
         headerBlock.root = bytes32(dataList[3].toUintStrict());
         headerBlock.createdAt = now;
     }
+
+    // HOUSEKEEPING functions
+    function setNextHeaderBlock(uint256 _value) public onlyOwner {
+        _nextHeaderBlock = _value;
+    }
+
+    function setHeimdallId(string memory _heimdallId) public onlyOwner {
```
0xAshish: how are we setting/updating `registry`?

0xAshish: before it was in the constructor.

0xAshish:
@atvanguard this should go to stage-1 branch

contracts/root/RootChain.sol
atvanguard: Missed committing `RootChainProxy`. Check now.

PR #180, opened 2020-02-14
Upgradable stakeManager

0xAshish:


contracts/staking/stakeManager/StakeManagerStorage.sol
```diff
@@ -0,0 +1,77 @@
...
+    // validator metadata
+    // mapping (uint256 => Validator) public validators;
+    //Mapping for epoch to totalStake for that epoch
+    mapping(uint256 => State) public validatorState;
+    //Ongoing auctions for validatorId
+    mapping(uint256 => Auction) public validatorAuction;
+
+    uint256 public totalHeimdallFee;
+
+    constructor() public Lockable(address(0x0)) {}
```
atvanguard: Storage shouldn't need to have a constructor.

contracts/staking/stakeManager/StakeManagerProxy.sol
```diff
@@ -0,0 +1,28 @@
...
+contract StakeManagerProxy is Proxy, StakeManagerStorage {
+    constructor(
+        address _proxyTo,
+        address _registry,
+        address _rootchain,
+        address _NFTContract,
+        address _stakingLogger,
+        address _validatorShareFactory,
+        address _governance
+    ) public Proxy(_proxyTo) {
```
atvanguard: `public Proxy(_proxyTo) Lockable(_governance)` because StakeManagerStorage inherits from Lockable.

contracts/staking/stakeManager/IStakeManager.sol
```diff
@@ -1,23 +1,8 @@
...
-        uint256 activationEpoch;
-        uint256 deactivationEpoch;
-        uint256 jailTime;
-        address signer;
-        address contractAddress;
-        Status status;
-    }
-
-    mapping(uint256 => Validator) public validators;
+contract IStakeManager is StakeManagerStorage {
```
atvanguard: Why does the interface need to inherit from storage?

contracts/staking/stakeManager/StakingNFT.sol
```diff
@@ -0,0 +1,34 @@
...
+contract StakingNFT is ERC721Full, Ownable {
+    constructor(string memory name, string memory symbol)
+        public
+        ERC721Full(name, symbol)
+    {
+        // solhint-disable-previous-line no-empty-blocks
+    }
+
+    function mint(address to, uint256 tokenId) public onlyOwner {
+        require(
```
atvanguard: IMO, this should be checked by client contract before calling this function.

contracts/staking/stakeManager/IStakeManager.sol
0xAshish: Using some variables as public functions and this interface as well. so all the contracts using `IStakeManager` can read public variables and use interface functions.

contracts/staking/stakeManager/StakingNFT.sol
0xAshish: `client contract` has no information regarding owned nft's so instead of calling 2 functions `balanceOf` and then `mint` it makes sense to have a single call.

contracts/staking/stakeManager/StakeManagerStorage.sol
0xAshish: removed

contracts/staking/stakeManager/StakeManagerProxy.sol
0xAshish: forgot that I can call 2 `constructor`, due to this one fixed above comment as well. Thanks 😄

contracts/staking/stakeManager/StakeManagerProxy.sol
```diff
@@ -0,0 +1,29 @@
...
+        address _stakingLogger,
+        address _validatorShareFactory,
+        address _governance
+    ) public Proxy(_proxyTo) Lockable(_governance) {
+        registry = _registry;
+        rootChain = _rootchain;
+        NFTContract = StakingNFT(_NFTContract);
+        logger = StakingInfo(_stakingLogger);
+        factory = ValidatorShareFactory(_validatorShareFactory);
+        governance = IGovernance(_governance);
```
atvanguard: This is not required because `Lockable` will do it.

0xAshish: yes, fixing

PR #189, opened 2020-02-19
Add housekeeping functions 2001->2002

atvanguard:


atvanguard:
Closed in favour of #190

PR #191, opened 2020-02-20
add drainable mechanism

nglglhtr:


contracts/common/misc/Drainable.sol
```diff
@@ -0,0 +1,38 @@
...
+import {IERC20} from "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
+import {IERC721} from "openzeppelin-solidity/contracts/token/ERC721/IERC721.sol";
+
+
+contract Drainable is Governable {
+
+  constructor (address _governance) public Governable(_governance) {}
+
+  function drainErc20 (
+    address[] calldata tokens,
```
atvanguard: Let's also accept the `uint256[] amount`. Better to be flexible. Same for the next 2 functions.

contracts/common/misc/Drainable.sol
```diff
@@ -0,0 +1,38 @@
+pragma solidity ^0.5.2;
+
+import { Governable } from "../governance/Governable.sol";
+
+import {IERC20} from "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
+import {IERC721} from "openzeppelin-solidity/contracts/token/ERC721/IERC721.sol";
+
+
+contract Drainable is Governable {
```
atvanguard: Inherit from `DepositManagerStorage` instead. Proxy upgradeable contracts need to have same memory layout.

contracts/common/misc/Drainable.sol
```diff
@@ -0,0 +1,33 @@
...
+
+contract Drainable is DepositManagerStorage {
+
+  function drainTokens(
+    address[] calldata tokens,
+    uint256[] calldata values,
+    address destination
+  ) external onlyGovernance {
+    for (uint256 i = 0; i < tokens.length; i++) {
+      if (registry.isERC721(tokens[i])) {
```
atvanguard: Let's remove this. Someone could send unmapped tokens to this contract as well. Have separate functions for draining 20/721s.

nglglhtr:
@atvanguard Would you suggest tracking the deposited ether in DepositManagerStorage contract? https://github.com/maticnetwork/contracts/blob/79b5a190be09929def4ac5a6ffa2aee27efa954b/contracts/root/depositManager/DepositManager.sol#L152, since the deposited value is directly transferred to MaticWETH contract

contracts/root/depositManager/DepositManager.sol
```diff
@@ -16,12 +16,13 @@ import {IDepositManager} from "./IDepositManager.sol";
...
 import {Lockable} from "../../common/mixin/Lockable.sol";
-
+import {Drainable} from "../../common/misc/Drainable.sol";
 contract DepositManager is
     DepositManagerStorage,
     IDepositManager,
     IERC721Receiver,
-    ContractReceiver
+    ContractReceiver,
+    Drainable
```
atvanguard: The entire point of this exercise was to replace the implementation of depositManager and then drain. Please don't inherit from `Drainable`. - In the test deploy a new instance of drainable, update address in depositManagerProxy and then call drain functions :)

test/helpers/deployer.js
```diff
@@ -169,6 +169,17 @@ class Deployer {
     return this.depositManager
   }
 
+  async deployDrainable() {
+    this.drainable = await contracts.Drainable.new()
+    await this.depositManagerProxy.updateImplementation(this.drainable.address)
+    await this.updateContractMap(
```
atvanguard: Why is this required?

test/root/Drainable.test.js
```diff
@@ -0,0 +1,80 @@
...
+      const values = []
+      tokens.push(testToken.address)
+      values.push(amount.toString())
+      drainable = await deployer.deployDrainable()
+      const data = drainable.contract.methods.drainErc20(tokens, values, destination).encodeABI()
+      await this.contracts.governance.update(
+        drainable.address,
+        data
+      )
+      utils.assertBigNumberEquality(await testToken.balanceOf(destination), amount)
```
atvanguard: Did you remove the assertion where balanceOf deposit Manager should now be 0. Also can we test for partial amount? Like if DepositManager has 10 we should drain only 7.

nglglhtr: yes, was that required? assertion for the user to get the queried amount of tokens should be enough, no?

test/helpers/deployer.js
nglglhtr: Just for consistency, isn't a requirement for this. Would you suggest removing it?

test/root/Drainable.test.js
nglglhtr: yes, will add for partial amount too

test/helpers/deployer.js
atvanguard: There is no question of consistency in contracts. Gotta save all storage and gas possible.

test/root/Drainable.test.js
atvanguard: > assertion for the user to get the queried amount of tokens should be enough, no
Isn't that just half the assertion? Better to write a complete story.

PR #192, opened 2020-02-20
Update StakeManager.sol

sanchaymittal:
Checking bid for higher amount before transferring token. Will considerably save the gas expenditure in case of a failure.


0xAshish:
Thanks @sanchaymittal

PR #194, opened 2020-02-23
start an auction in the start for new validators

0xAshish:
- Start auction in the beginning

### DIFF
--- a/contracts/staking/stakeManager/StakeManager.sol
+++ b/contracts/staking/stakeManager/StakeManager.sol
@@ -102,23 +102,18 @@ contract StakeManager is IStakeManager {
             replacementCoolDown == 0 || replacementCoolDown <= currentEpoch,
             "Cool down period"
         );
-        require(
-            auctionPeriod >=
-                currentEpoch.sub(
-                    validatorAuction[validatorId].startEpoch.add(dynasty)
-                ),
-            "Invalid auction period"
-        );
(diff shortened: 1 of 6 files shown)


contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -102,23 +102,18 @@ contract StakeManager is IStakeManager {
...
+        // (auctionPeriod--dynasty)--(auctionPeriod--dynasty)--(auctionPeriod--dynasty)
+        // if it's auctionPeriod then will get residue smaller then auctionPeriod
+        // from (CurrentPeriod of validator )%(auctionPeriod--dynasty)
         // make sure that its `auctionPeriod` window
         // dynasty = 30, auctionPeriod = 7, activationEpoch = 1, currentEpoch = 39
-        // residue 1 = (39-1)% (30+7), if residue-dynasty  > 0 it's `auctionPeriod`
+        // residue 1 = (39-1)% (7+30), if residue <= auctionPeriod it's `auctionPeriod`
+
         require(
             (currentEpoch.sub(validators[validatorId].activationEpoch) %
```
atvanguard: How about we make it independent of `activationEpoch`? It will be easier to know that in an auction period, it's possible to outbid any validator; whereas here, auction periods for each validator will vary.

0xAshish: it is intentionally kept this way to avoid sudden replacement of the majority validators.

PR #195, opened 2020-02-24
make stakeManager owner for validator share contracts

0xAshish:


contracts/staking/validatorShare/ValidatorShareFactory.sol
```diff
@@ -7,16 +7,14 @@ contract ValidatorShareFactory {
...
+        returns (address)
+    {
         ValidatorShare validatorShare = new ValidatorShare(
             validatorId,
             loggerAddress,
-            msg.sender,
-            governance
+            msg.sender
         );
         validatorShare.transferOwnership(msg.sender);
```
atvanguard: Instead of this, we could call `_transferOwnership` in `validatorShare` constructor since we are passing the `stakeManager` address now as a param in the constructor.

0xAshish: ah yes fixed

PR #209, opened 2020-03-19
add reward param into staking info for external reading

0xAshish:


jdkanani:
@0xAshish @atvanguard Please **do not push** these changes in cs-2004.

contracts/staking/StakingInfo.sol
```diff
@@ -257,6 +258,7 @@ contract StakingInfo {
         view
         returns (
             uint256 amount,
+            uint256 reward,
```
atvanguard: If you are adding a return value, then I'd expect this to be used somewhere but I don't see an equivalent change in this PR?

0xAshish: oh on Heimdall? no, that pr is fixing these changes.
Since stakingInfo is helper contract this function is meant for things like staking-API and other dapps interacting with it.

Issue #212, opened 2020-03-23
MAT-918 Incorrect block interval


jdkanani:
Block interval should `end-start+1` at https://github.com/maticnetwork/contracts/blob/develop/contracts/root/RootChain.sol#L56

Reviewers: @atvanguard @jdkanani

PR #213, opened 2020-03-24
fix block interval calculation

0xAshish:
fix for https://github.com/maticnetwork/contracts/issues/212

### DIFF
--- a/contracts/root/RootChain.sol
+++ b/contracts/root/RootChain.sol
@@ -53,7 +53,7 @@ contract RootChain is RootChainStorage, IRootChain {
         );
         // blockInterval, voteHash, stateRoot, sigs
         uint256 _reward = stakeManager.checkSignatures(
-            headerBlock.end.sub(headerBlock.start),
+            headerBlock.end.sub(headerBlock.start.add(1)),
             keccak256(vote),
             bytes32(extraData[4].toUint()),
             sigs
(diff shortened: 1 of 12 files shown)


contracts/root/RootChain.sol
```diff
@@ -53,7 +53,7 @@ contract RootChain is RootChainStorage, IRootChain {
         );
         // blockInterval, voteHash, stateRoot, sigs
         uint256 _reward = stakeManager.checkSignatures(
-            headerBlock.end.sub(headerBlock.start),
+            headerBlock.end.sub(headerBlock.start.add(1)),
```
jdkanani:
```suggestion
            headerBlock.end.sub(headerBlock.start).add(1),
```

0xAshish: Thanks fixed.

contracts/staking/StakingInfo.sol
```diff
@@ -99,6 +99,11 @@ contract StakingInfo {
         uint256 indexed rewards,
         uint256 tokens
     );
+    event DelReStaked(
```
atvanguard: Not clear from the name what this event does.

0xAshish: Del is a prefix for delegation related events and `ReStaked` is for `reStaking`.

atvanguard: Clearer name would have helped. Approving...

PR #215, opened 2020-03-30
Bug fix: sell voucher

0xAshish:
- Amount was subtracted from the `activeAmount` which wasn't the correct way because `activeAmount` doesn't include the rewards amount.
- `activeAmount` stores amount locked by delegators, `_amount` is the total amount for one delegator (his `activeAmount`+ his `rewards`.

### DIFF
--- a/contracts/staking/validatorShare/ValidatorShare.sol
+++ b/contracts/staking/validatorShare/ValidatorShare.sol
@@ -212,6 +212,7 @@ contract ValidatorShare is IValidatorShare {
             ),
             "Insufficent rewards"
         );
+        stakingLogger.logDelUnstaked(validatorId, msg.sender, delegator.amount);
         delete delegators[msg.sender];
     }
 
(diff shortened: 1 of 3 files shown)


contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -131,6 +132,7 @@ contract ValidatorShare is IValidatorShare {
 
         stakingLogger.logShareBurned(validatorId, msg.sender, _amount, share);
         stakingLogger.logStakeUpdate(validatorId);
+
```
jdkanani: Remove unwanted extra line.

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -121,6 +120,8 @@ contract ValidatorShare is IValidatorShare {
             _amount = _amount.sub(_rewards);
         }
 
+        activeAmount = activeAmount.sub(_amount);
```
jdkanani: Please explain what changed in `activeAmount` before and after this change?

It would be great if we can have scenarios in PR comments where it was breaking and where it was working, to track the timeline of the problem.

jdkanani:
@0xAshish Please merge develop branch to avoid conflict.

PR #216, opened 2020-03-31
Instead of signer address use signer pubkey

0xAshish:
- Heimdall requires signer `pubkey` for validator join tx so adding it into the staked event.
- validator struct still stores key extracted from `pubkey`
- signer change also emits an event with `pubkey` so that Heimdall can listen and add that key.

### DIFF
--- a/contracts/staking/StakingInfo.sol
+++ b/contracts/staking/StakingInfo.sol
@@ -25,11 +25,11 @@ contract IStakeManager {
 
 contract StakingInfo {
     event Staked(
-        address indexed signer,
         uint256 indexed validatorId,
         uint256 indexed activationEpoch,
         uint256 amount,
-        uint256 total
+        uint256 total,
+        bytes signerPubkey
     );
     event Unstaked(
         address indexed user,
@@ -48,7 +48,7 @@ contract StakingInfo {
     event SignerChange(
         uint256 indexed validatorId,
         address indexed oldSigner,
-        address indexed newSigner
+        bytes newSignerPubkey
     );
     event ReStaked(uint256 indexed validatorId, uint256 amount, uint256 total);
     event Jailed(uint256 indexed validatorId, uint256 indexed exitEpoch);
@@ -143,13 +143,13 @@ contract StakingInfo {
     }
 
     function logStaked(
-        address signer,
+        bytes memory signerPubkey,
         uint256 validatorId,
         uint256 activationEpoch,
         uint256 amount,
         uint256 total
     ) public onlyStakeManager {
-        emit Staked(signer, validatorId, activationEpoch, amount, total);
+        emit Staked(validatorId, activationEpoch, amount, total, signerPubkey);
     }
 
     function logUnstaked(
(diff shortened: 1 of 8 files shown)


contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -720,4 +739,11 @@ contract StakeManager is IStakeManager {
         validatorState[epoch].stakerCount += stakerCount;
     }
 
+    function pubToAddress(bytes memory pub) public pure returns (address addr) {
+        bytes32 hash = keccak256(pub);
+        assembly {
+            mstore(0, hash)
```
atvanguard, line 745: Shouldn't we use the free memory pointer at `0x40` ?

jdkanani: We must check if the `pub`'s length is 64 otherwise it will give a different address accidently.

contracts/staking/StakingInfo.sol
```diff
@@ -143,13 +143,13 @@ contract StakingInfo {
...
     function logStaked(
-        address signer,
+        bytes memory signerPubkey,
         uint256 validatorId,
         uint256 activationEpoch,
         uint256 amount,
         uint256 total
     ) public onlyStakeManager {
-        emit Staked(signer, validatorId, activationEpoch, amount, total);
+        emit Staked(validatorId, activationEpoch, amount, total, signerPubkey);
```
jdkanani, line 152: Having a signer in the event also gives flexibility to UI and data processors like Hermione.

contracts/staking/StakingInfo.sol
```diff
@@ -173,9 +173,9 @@ contract StakingInfo {
     function logSignerChange(
         uint256 validatorId,
         address oldSigner,
-        address newSigner
+        bytes memory signerPubkey
     ) public onlyStakeManager {
-        emit SignerChange(validatorId, oldSigner, newSigner);
+        emit SignerChange(validatorId, oldSigner, signerPubkey);
```
jdkanani, line 178: Would be great if we could add indexed-signer as well in the event for the easier query.

contracts/staking/stakeManager/StakeManager.sol
jdkanani: We must check `pub`'s length to 64 to avoid accidental address as signer.

mankenavenkatesh:
Pls change stake.js accordingly for stake and signerchange calls.

contracts/staking/StakingInfo.sol
0xAshish: Added

contracts/staking/stakeManager/StakeManager.sol
atvanguard: > Shouldn't we use the free memory pointer at 0x40 ?

Might not be required. Solidity docs say that `0x00 - 0x3f (64 bytes): scratch space for hashing methods`. Not sure what it means, please verify :) https://solidity.readthedocs.io/en/v0.5.12/miscellaneous.html#layout-in-memory

contracts/staking/StakingInfo.sol
0xAshish: Done

contracts/staking/stakeManager/StakeManager.sol
0xAshish: yes saw that, and went back to type casting.

PR #217, opened 2020-04-02
Confirm auction logic update

0xAshish:
- The new validator can call `confirmAuction` anytime in next dynasty
- After which auction continues as intended
- If there were no bids the previous bidder can call `confirmAuction`.


jdkanani:
@0xAshish CI is failing.

PR #219, opened 2020-04-06
remove todos, add cool down for commission rate updates

0xAshish:


contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -304,12 +304,12 @@ contract StakeManager is IStakeManager {
         );
     }
 
+    /// @notice currently replace with just minimum amount
```
atvanguard, line 307: Not clear from the comment what the function was intended for. I'd suggest simply removing the function for now for clarity.

contracts/staking/SlashingManager.sol
```diff
@@ -31,7 +32,6 @@ contract SlashingManager is Ownable {
         bytes memory sig1,
         bytes memory sig2
     ) public {
-        // Todo: fix signer chanage for same validator
```
atvanguard, line 34: So, is this fixed?

contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -404,79 +403,7 @@ contract StakeManager is IStakeManager {
...
-            );
-            ValidatorShare(validators[validatorId].contractAddress).lock();
-        }
-        // update future in case of no `unJail`
-        updateTimeLine(exitEpoch, -(int256(amount) + delegationAmount), -1);
-        validators[validatorId].deactivationEpoch = exitEpoch;
-        validators[validatorId].status = Status.Locked;
-        logger.logJailed(validatorId, exitEpoch);
-    }
+    ) public onlySlashingMananger {}
```
atvanguard, line 406: Ummmm unimplemented function. Having an empty function around creates confusion. I'd suggest removing unimplemented functions.

contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -588,7 +515,7 @@ contract StakeManager is IStakeManager {
                 validators[validatorId].activationEpoch <= currentEpoch) &&
             (validators[validatorId].deactivationEpoch == 0 ||
                 validators[validatorId].deactivationEpoch > currentEpoch) &&
-            validators[validatorId].status == Status.Active); // Todo: reduce logic
```
atvanguard, line 591: Do you think we should actually reduce the logic instead of todos?

contracts/staking/SlashingManager.sol
```diff
@@ -73,5 +73,4 @@ contract SlashingManager is Ownable {
     }
 
     function checkpointHalt(uint256 start) public {}
```
atvanguard, line 75: We can add this function when it is actually implemented. Thoughts?

contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -621,7 +548,6 @@ contract StakeManager is IStakeManager {
     ) internal returns (uint256) {
         // total voting power
         uint256 _stakePower;
-        //TODO: add proposer bonus
```
atvanguard, line 624: Not sure what "proposer bonus" is, but did we add it, since we are removing the todo?

contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -640,7 +566,6 @@ contract StakeManager is IStakeManager {
                     valPow = ValidatorShare(validator.contractAddress)
                         .udpateRewards(_reward, stakePower);
                 } else {
-                    //TODO: check for div leaks in rewards amount
```
atvanguard, line 643: So are the leaks being accounted for?

contracts/staking/stakeManager/StakeManagerStorage.sol
```diff
@@ -33,7 +34,7 @@ contract StakeManagerStorage is ProxyStorage, Lockable, RootChainable {
     uint256 public auctionPeriod = dynasty / 4; // 1 week in epochs
     bytes32 public accountStateRoot;
 
-    // on dynasty update certain amount of cooldown period where there is no validator auction
+    // on dynasty update for a certain amount of cooldown period where there is no validator auction
```
atvanguard, line 37: Not clear to me what this comment means.

contracts/staking/validatorShare/IValidatorShare.sol
```diff
@@ -11,13 +11,18 @@ import {Lockable} from "../../common/mixin/Lockable.sol";
...
 
+
 contract IValidatorShare is ERC20, Lockable, Ownable {
     using SafeMath for uint256;
     StakingInfo public stakingLogger;
     IStakeManager public stakeManager;
     uint256 public validatorId;
     uint256 public validatorRewards;
     uint256 public commissionRate;
+    ///@notice # of checkpoints after which commission rate can be updated
```
atvanguard, line 22: I think both `///` and `@notice` is meant for class, interface and functions.

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -4,10 +4,8 @@ import {Registry} from "../../common/Registry.sol";
 
 import {IValidatorShare} from "./IValidatorShare.sol";
 
-// TODO: refactor each function to reusable smaller internal functions
+
 contract ValidatorShare is IValidatorShare {
-    // TODO: totalStake and active ammount issue
```
atvanguard, line 9: Not sure whether we should be resolving these todos or removing them? 3 others in this file.

contracts/staking/validatorShare/IValidatorShare.sol
0xAshish: https://solidity.readthedocs.io/en/v0.5.10/natspec-format.html
Not sure what to use for variables. `if no tags are used then the Solidity compiler will interpret a /// or /** comment in the same way as if it were tagged with @notice.` so using notice makes sense.

contracts/staking/validatorShare/ValidatorShare.sol
0xAshish: this one is resolved, `totalStake` is not used for any real accounting for rewards etc.used to keep track of total stake locked in the contract for external reads

contracts/staking/stakeManager/StakeManager.sol
0xAshish: These are minor leaks for smaller decimal points. Adding it into linear issue, as of now it doesn't seem to be problematic.

contracts/staking/stakeManager/StakeManager.sol
0xAshish: no, as of now choose to not give any bonus to proposers.

contracts/staking/SlashingManager.sol
0xAshish: removed, FYI slashing is WIP.

0xAshish:
@atvanguard removed `slashingManager` and resolved related comments. Slashing in new PR.

contracts/staking/stakeManager/StakeManagerStorage.sol
0xAshish: fixed

contracts/staking/stakeManager/StakeManager.sol
0xAshish: tried to do it but no luck, depending on slashing we may be able to remove status var.

contracts/staking/stakeManager/StakingNFT.sol
```diff
@@ -16,7 +17,7 @@ contract StakingNFT is ERC721Full, Ownable {
             balanceOf(to) == 0,
             "Validators MUST NOT own multiple stake position"
         );
-        //require(balanceOf(user) == 0, "Only one time staking is allowed");
+        require(balanceOf(to) == 0, "Only one time staking is allowed");
```
arthcp, line 20: Aren't both these require conditions same?

0xAshish: oh lol @arthcp Thanks

contracts/staking/stakeManager/StakeManager.sol
0xAshish: this comment is not relevant since mechanism/code is updated.

PR #220, opened 2020-04-08
Slashing

0xAshish:


contracts/common/Registry.sol
```diff
@@ -126,6 +127,10 @@ contract Registry is Governable {
         return contractMap[STAKE_MANAGER];
     }
 
+    function getSlashingManagerAddress() public view returns (address) {
```
coinsandsteeldev, line 130: From i've seen in the code - some getters follow solidity way - without using "get" and some code uses "get". I don't have preferences here, but it would be awesome if the code follows same rules.

contracts/staking/stakeManager/IStakeManager.sol
```diff
@@ -2,17 +2,19 @@ pragma solidity ^0.5.2;
...
+
     function confirmAuctionBid(
         uint256 validatorId,
         uint256 heimdallFee,
         bool acceptDelegation,
         bytes calldata signerPubkey
     ) external;
 
-    function delegationTransfer(
+    function transferfunds(
```
coinsandsteeldev, line 17: Small typo here - transferFunds

0xAshish: ah! yes thanks

contracts/common/Registry.sol
0xAshish: yes, I'll refactor it in a separate PR.

contracts/staking/StakingInfo.sol
```diff
@@ -219,11 +226,22 @@ contract StakingInfo {
...
+    function logUnJailed(uint256 validatorId, address signer)
         public
         onlyStakeManager
     {
-        emit Jailed(validatorId, exitEpoch);
+        emit UnJailed(validatorId, signer);
+    }
+
+    function logSlashed(uint256 amount) public onlyStakeManager {
+        emit Slashed(amount);
```
atvanguard, line 244: Should we also emit who was slashed?

contracts/staking/slashing/ISlashingManager.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity ^0.5.2;
+import {StakingInfo} from "../StakingInfo.sol";
+import {Registry} from "../../common/Registry.sol";
+
+
+contract ISlashingManager {
```
atvanguard, line 6: I don't think there is a need to initialize storage variables to a specific value here.
Might be a good idea to keep storage and interface separate.

contracts/staking/slashing/SlashingManager.sol
```diff
@@ -0,0 +1,126 @@
...
+    using SafeMath for uint256;
+    using RLPReader for bytes;
+    using RLPReader for RLPReader.RLPItem;
+
+    modifier onlyStakeManager() {
+        require(registry.getStakeManagerAddress() == msg.sender);
+        _;
+    }
+
+    constructor(address _registry, address _logger) public {
```
atvanguard, line 24: Please accept `_heimdallId` as an initialization param.

contracts/staking/slashing/SlashingManager.sol
```diff
@@ -0,0 +1,126 @@
...
+
+        require(
+            keccak256(dataList[4].toBytes()) ==
+                keccak256(abi.encodePacked(sha256(txData))),
+            "Extra data is invalid"
+        );
+
+        uint256 stakePower;
+        uint256 activeTwoByThree;
+        (stakePower, activeTwoByThree) = logger.verifyConsensus(
```
atvanguard, line 56: logger being used to verify consensus sounds counter-intuitive.

contracts/staking/slashing/SlashingManager.sol
```diff
@@ -0,0 +1,126 @@
...
+            keccak256(dataList[1].toBytes()) ==
+                keccak256(abi.encodePacked(sha256(slashingInfoList))),
+            "Invalid slashInfoHash"
+        );
+        //slashingInfoList[]=[[valiD,am,isJailed]]
+        uint256 slashedAmount = stakeManager.slash(slashingInfoList);
+        uint256 bounty = (slashedAmount.mul(reportRate)).div(100);
+        slashedAmount = slashedAmount.sub(bounty);
+        require(
+            stakeManager.transferFunds(
```
atvanguard, line 73: To avoid duplication, might be a good idea to place `stakeManager.transferFunds` in an internal function and call that everywhere.

contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -645,9 +742,8 @@ contract StakeManager is IStakeManager, StakeManagerStorage {
             ValidatorShare validatorShare = ValidatorShare(
                 validators[validatorId].contractAddress
             );
-            delegationAmount = int256(validatorShare.activeAmount());
             rewards = rewards.add(validatorShare.withdrawRewardsValidator());
-            validatorShare.lock();
+            delegationAmount = int256(validatorShare.lockContract());
```
atvanguard, line 746: lockContract() returning an int256 sounds bad naming.

0xAshish: trying to reduce internal calls. There are some other changes that need lock/unlock, will change naming and design there only. (depending on those requirements)

contracts/staking/StakingInfo.sol
0xAshish: no need list of validators is already in `calldata`.

contracts/staking/slashing/SlashingManager.sol
0xAshish: reading from root chain is better 🤔💭.

contracts/staking/slashing/SlashingManager.sol
0xAshish: sorry I didn't get😐

contracts/staking/slashing/SlashingManager.sol
0xAshish: True, just reusing code.

contracts/staking/slashing/SlashingManager.sol
atvanguard: You are not reading from RootChain either, you are inheriting a hardcoded value from `ISlashingManager`.

contracts/staking/slashing/SlashingManager.sol
atvanguard: I'd suggest refactoring the code so that logger only does what it should.

contracts/staking/slashing/SlashingManager.sol
atvanguard:
```
function _stakeFundsTransfer() {
require(
            stakeManager.transferFunds(0,
                slashedAmount,
                address(this)
            ),
            "Transfer failed"
        );
```
And then use `_stakeFundsTransfer`, everywhere.

0xAshish: sorry I still don't get the point because I would still need to call `_stakeFundsTransfer` 3 times with different params. which will do the same thing as the current one eventually.

contracts/staking/slashing/SlashingManager.sol
0xAshish: taking a look.

contracts/staking/slashing/SlashingManager.sol
0xAshish: oh💩 lost that change, added into the initialization param.

contracts/staking/slashing/SlashingManager.sol
0xAshish: refactored

contracts/staking/slashing/ISlashingManager.sol
0xAshish: done

PR #222, opened 2020-04-10
Add checkpoint verify

0xAshish:


contracts/staking/StakingInfo.sol
```diff
@@ -21,9 +26,18 @@ contract IStakeManager {
     bytes32 public accountStateRoot;
     uint256 public activeAmount; // delegation amount from validator contract
     uint256 public validatorRewards;
+    // signer to Validator mapping
+    mapping(address => uint256) public signerToValidator;
```
jdkanani, line 30: Shouldn't this be a function?  Why need to use storage for the same?

0xAshish: this is just a dummy interface, but yes one function will also do. Thanks

PR #231, opened 2020-04-29
change: side-channel: submit header block

AyushKaul:
- [x] fix deloyment
- [x] test submitBlockHeader()
- [x] change tests

### DIFF
--- a/contracts/root/RootChain.sol
+++ b/contracts/root/RootChain.sol
@@ -10,6 +10,7 @@ import {IRootChain} from "./IRootChain.sol";
 import {Registry} from "../common/Registry.sol";
 import "../common/lib/BytesLib.sol";
 
+
 contract RootChain is RootChainStorage, IRootChain {
     using SafeMath for uint256;
     using RLPReader for bytes;
@@ -72,46 +73,48 @@ contract RootChain is RootChainStorage, IRootChain {
     //     _blockDepositId = 1;
     // }
 
-    // side channel 
-    function submitHeaderBlock(
-        bytes calldata data,
-        bytes calldata sigs
-    ) external {
-        (address proposer,
-        uint256 start,
-        uint256 end,
-        bytes32 rootHash,
-        bytes32 accountHash,
-        bytes32 _borChainID) = 
-        abi.decode(data, (address, uint256, uint256, bytes32, bytes32, bytes32));
+    // side channel
+    function submitHeaderBlock(bytes calldata data, bytes calldata sigs)
+        external
+    {
+        (
+            address proposer,
+            uint256 start,
+            uint256 end,
+            bytes32 rootHash,
+            bytes32 accountHash,
+            bytes32 _borChainID
+        ) = abi.decode(
+            data,
+            (address, uint256, uint256, bytes32, bytes32, bytes32)
+        );
         // require(bytes32(uint(15001)) == _borChainID, "Invalid bor chain id");
 
-        RootChainHeader.HeaderBlock memory headerBlock = _buildHeaderBlock(
-            proposer, start, end, rootHash
+        require(
+            _buildHeaderBlock(proposer, start, end, rootHash),
+            "INCORRECT_HEADER_DATA"
         );
-        headerBlocks[_nextHeaderBlock] = headerBlock;
+
         // check if it is better to keep it in local storage instead
         IStakeManager stakeManager = IStakeManager(
             registry.getStakeManagerAddress()
         );
 
-        bytes memory votePrefix = hex'01';
-        
         // blockInterval, voteHash, stateRoot, sigs
         uint256 _reward = stakeManager.checkSignatures(
-            headerBlock.end.sub(headerBlock.start).add(1),
-            keccak256(BytesLib.concat(votePrefix, data)), // prefix 01 to data
+            end.sub(start).add(1),
+            keccak256(abi.encodePacked(bytes(hex"01"), data)), // prefix 01 to data
             accountHash,
             sigs
         );
         require(_reward != 0, "Invalid checkpoint");
         emit NewHeaderBlock(
-            headerBlock.proposer,
+            proposer,
             _nextHeaderBlock,
             _reward,
-            headerBlock.start,
-            headerBlock.end,
-            headerBlock.root
+            start,
+            end,
+            rootHash
         );
         _nextHeaderBlock = _nextHeaderBlock.add(MAX_DEPOSITS);
         _blockDepositId = 1;
@@ -144,14 +147,13 @@ contract RootChain is RootChainStorage, IRootChain {
         return _nextHeaderBlock.sub(MAX_DEPOSITS);
     }
 
-    function _buildHeaderBlock(address proposer,
-    uint start,
-    uint end,
-    bytes32 rootHash)
-        private
-        view
-        returns (HeaderBlock memory headerBlock)
-    {
+    function _buildHeaderBlock(
+        address proposer,
+        uint256 start,
+        uint256 end,
+        bytes32 rootHash
+    ) private returns (bool) {
+        HeaderBlock memory headerBlock;
         headerBlock.proposer = proposer;
         // Is this required? Why does a proposer need to be the sender? Think validator relay networks
         // require(msg.sender == dataList[0].toAddress(), "Invalid proposer");
@@ -164,16 +166,18 @@ contract RootChain is RootChainStorage, IRootChain {
         if (_nextHeaderBlock > MAX_DEPOSITS) {
             nextChildBlock = headerBlocks[currentHeaderBlock()].end + 1;
         }
-        require(
-            nextChildBlock == start,
-            "INCORRECT_START_BLOCK"
-        );
+        // require(nextChildBlock == start, "INCORRECT_START_BLOCK");
+        if (nextChildBlock != start) {
+            return false;
+        }
         headerBlock.start = nextChildBlock;
         headerBlock.end = end;
 
         // toUintStrict returns the encoded uint. Encoded data must be padded to 32 bytes.
         headerBlock.root = rootHash;
         headerBlock.createdAt = now;
+        headerBlocks[_nextHeaderBlock] = headerBlock;
+        return true;
     }
 
     // Housekeeping function. @todo remove later
--- a/contracts/root/RootChain.sol
+++ b/contracts/root/RootChain.sol
@@ -88,7 +88,7 @@ contract RootChain is RootChainStorage, IRootChain {
             data,
             (address, uint256, uint256, bytes32, bytes32, bytes32)
         );
-        // require(bytes32(uint(15001)) == _borChainID, "Invalid bor chain id");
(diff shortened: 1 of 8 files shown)


contracts/root/RootChain.sol
```diff
@@ -22,50 +24,97 @@ contract RootChain is RootChainStorage, IRootChain {
...
         );
+
         // blockInterval, voteHash, stateRoot, sigs
         uint256 _reward = stakeManager.checkSignatures(
-            headerBlock.end.sub(headerBlock.start).add(1),
-            keccak256(vote),
-            bytes32(extraData[4].toUint()),
+            end.sub(start).add(1),
+            keccak256(abi.encodePacked(bytes(hex"01"), data)), // prefix 01 to data
+            accountHash,
```
jdkanani, line 107: @0xAshish This should be root, not accountHash

contracts/root/RootChain.sol
```diff
@@ -22,50 +24,97 @@ contract RootChain is RootChainStorage, IRootChain {
...
+            uint256 end,
+            bytes32 rootHash,
+            bytes32 accountHash,
+            bytes32 _borChainID
+        ) = abi.decode(
+            data,
+            (address, uint256, uint256, bytes32, bytes32, bytes32)
         );
-        require(dataList[1].toUint() == VOTE_TYPE, "Vote type is invalid");
+        // require(bytes32(uint(15001)) == _borChainID, "Invalid bor chain id");
```
AyushKaul, line 91: uncomment

contracts/root/RootChain.sol
0xAshish: seems correct to me

contracts/root/RootChain.sol
```diff
@@ -88,7 +88,7 @@ contract RootChain is RootChainStorage, IRootChain {
             data,
             (address, uint256, uint256, bytes32, bytes32, bytes32)
         );
-        // require(bytes32(uint(15001)) == _borChainID, "Invalid bor chain id");
+        require(bytes32(uint256(15001)) == _borChainID, "Invalid bor chain id");
```
AyushKaul, line 91: @0xAshish Is there a global constant for borChainID?

AyushKaul: done

jdkanani:
@0xAshish @AyushKaul Created new branch called `sidechannel` from `develop` and changed base for this PR to `sidechannel`

jdkanani:
Looks good!

test/helpers/utils.js
```diff
@@ -92,34 +92,24 @@ export function buildSubmitHeaderBlockPaylod(
...
-      2,
-      0,
-      0,
-      ethUtils.bufferToHex(ethUtils.sha256(extraData))
-    ])
+  let data = web3.eth.abi.encodeParameters(
+    ['address', 'uint256', 'uint256', 'bytes32', 'bytes32', 'bytes32'],
+    [proposer, start, end, root, options.rewardsRootHash, '0x0000000000000000000000000000000000000000000000000000000000003a99']
   )
+  const sigData = Buffer.concat([ethUtils.toBuffer('0x01'), ethUtils.toBuffer(data)])
```
atvanguard, line 104: We are also appending 01 in solidity right? Why do it here before sending in the data? I might be confused.

test/helpers/utils.js
```diff
@@ -92,34 +92,24 @@ export function buildSubmitHeaderBlockPaylod(
...
 
   // in case of TestStakeManger use dummysig data
   const sigs = ethUtils.bufferToHex(
     options.getSigs
-      ? encodeSigs(getSigs(validators, ethUtils.keccak256(vote)))
+      ? encodeSigs(getSigs(validators, ethUtils.keccak256(sigData)))
       : 'dummySig'
   )
-  return { vote, sigs, extraData, root }
+  return { data, sigs, root }
```
atvanguard, line 112: Is the `root` required anymore?

test/helpers/utils.js
0xAshish: @jdkanani can you please comment.

contracts/root/RootChain.sol
```diff
@@ -98,12 +95,12 @@ contract RootChain is RootChainStorage, IRootChain {
...
-        returns (HeaderBlock memory headerBlock)
-    {
-        headerBlock.proposer = dataList[0].toAddress();
+    function _buildHeaderBlock(
+        address proposer,
+        uint256 start,
+        uint256 end,
+        bytes32 rootHash
+    ) private returns (bool) {
         // Is this required? Why does a proposer need to be the sender? Think validator relay networks
```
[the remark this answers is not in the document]
0xAshish: nope, this is not required if we go with the proposer bonus then also we can use it in staking functions.

contracts/root/RootChain.sol
```diff
@@ -22,50 +23,46 @@ contract RootChain is RootChainStorage, IRootChain {
...
         IStakeManager stakeManager = IStakeManager(
             registry.getStakeManagerAddress()
         );
         // blockInterval, voteHash, stateRoot, sigs
         uint256 _reward = stakeManager.checkSignatures(
-            headerBlock.end.sub(headerBlock.start).add(1),
-            keccak256(vote),
-            bytes32(extraData[4].toUint()),
+            end.sub(start).add(1),
+            keccak256(abi.encodePacked(bytes(hex"01"), data)), // prefix 01 to data
```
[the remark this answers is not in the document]
0xAshish: AFAIK validators can send data where prefix is 00 which means Noack(basically no vote but signed). so this is required to add inside the contract only. One thing we can do is just check for prefix instead of adding.

contracts/root/RootChain.sol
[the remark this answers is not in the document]
0xAshish: cc @jdkanani

test/helpers/utils.js
0xAshish: in a few tests we are using is in assertions.

atvanguard: This function takes root as a parameter. Please make the assertions on root before calling it. And let's remove it from the return value.

contracts/root/RootChain.sol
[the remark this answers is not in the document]
atvanguard: Yeah I just asked to write the reason in solidity code as a comment.

contracts/root/RootChain.sol
```diff
@@ -8,8 +8,9 @@ import {RootChainHeader, RootChainStorage} from "./RootChainStorage.sol";
 import {IStakeManager} from "../staking/stakeManager/IStakeManager.sol";
 import {IRootChain} from "./IRootChain.sol";
 import {Registry} from "../common/Registry.sol";
+import {ChainIdMixin} from "../common/mixin/ChainIdMixin.sol";
 
-contract RootChain is RootChainStorage, IRootChain {
+contract RootChain is RootChainStorage, IRootChain, ChainIdMixin {
```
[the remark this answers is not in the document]
AyushKaul: done

test/helpers/StatefulUtils.js
```diff
@@ -2,12 +2,7 @@ import ethUtils from 'ethereumjs-util'
 
 import { getBlockHeader } from './blocks'
 import MerkleTree from './merkle-tree'
-import {
```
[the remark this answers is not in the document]
AyushKaul: done

test/helpers/utils.js
0xAshish: Done

contracts/root/RootChain.sol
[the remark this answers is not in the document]
0xAshish: removed.

contracts/root/RootChain.sol
```diff
@@ -115,16 +114,22 @@ contract RootChain is RootChainStorage, IRootChain {
...
-        headerBlock.start = nextChildBlock;
-        headerBlock.end = dataList[2].toUint();
+        if (nextChildBlock != start) {
+            return false;
+        }
 
         // toUintStrict returns the encoded uint. Encoded data must be padded to 32 bytes.
-        headerBlock.root = bytes32(dataList[3].toUintStrict());
-        headerBlock.createdAt = now;
+        // headerBlock.root = bytes32(dataList[3].toUintStrict());
```
atvanguard, line 122: We can remove lines 121/122 now.

test/root/RootChain.test.js
```diff
@@ -62,24 +64,43 @@ contract('RootChain', async function (accounts) {
...
+    const { data, sigs } = buildSubmitHeaderBlockPaylod(
+      accounts[0],
+      0,
+      255,
+      '' /* root */,
+      wallets,
+      { allValidators: true, rewardsRootHash: tree.getRoot(), getSigs: true, totalStake: totalStake, sigPrefix: '0x00' }
+    )
+    try {
+      await rootChain.submitHeaderBlock(data, sigs)
```
atvanguard, line 79: Please add `assert.fail('Expected the tx to revert')`, otherwise the test will pass even if the tx doesn't revert.

test/root/RootChain.test.js
```diff
@@ -62,24 +64,43 @@ contract('RootChain', async function (accounts) {
...
+    } catch (error) {
+      const invalidOpcode = error.message.search('revert') >= 0
```
atvanguard, line 81: Can we assert on the actual expected reverted msg? Otherwise if it fails for some other reason the test won't detect it.

test/root/RootChain.test.js
```diff
@@ -62,24 +64,43 @@ contract('RootChain', async function (accounts) {
...
     )
-    const result = await rootChain.submitHeaderBlock(vote, sigs, extraData)
+    const result = await rootChain.submitHeaderBlock(data, sigs)
     const logs = result.logs
     logs.should.have.lengthOf(1)
     logs[0].event.should.equal('NewHeaderBlock')
     expect(logs[0].args).to.include({
       proposer: accounts[0],
-      root: '0x' + root.toString('hex')
+      root: ethUtils.bufferToHex(ethUtils.keccak256(encode(0, 255)))
```
atvanguard, line 103: This is just a dummy root assertion right? can we remove it?

test/root/RootChain.test.js
0xAshish: revert comes from `ecrecover` so no error msg. I think we can add some msg there and assert the same here.

test/root/RootChain.test.js
0xAshish: why remove? this is correct only.

test/staking/StakeManager.test.js
```diff
@@ -715,7 +715,7 @@ contract('StakeManager: Heimdall fee', async function (accounts) {
       // 2/3 majority vote
       await stakeManager.checkSignatures(
         22,
-        utils.bufferToHex(utils.keccak256(vote)),
+        utils.bufferToHex(utils.keccak256(Buffer.concat([utils.toBuffer('0x01'), utils.toBuffer(data)]))),
```
jdkanani, line 718: We should add more test cases with 0x02, non-majority vote to test multiple branches.

PR #234, opened 2020-05-05
Refactor and improve tests

imyourm8:


contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -32,22 +32,21 @@ contract ValidatorShare is IValidatorShare {
...
-                _rewards.sub(_valRewards).mul(commissionRate).div(100)
+            _validatorRewards = _validatorRewards.add(
+                _rewards.sub(_validatorRewards).mul(commissionRate).div(100)
             );
         }
-        _rewards = _rewards.sub(_valRewards);
-        validatorRewards = validatorRewards.add(_valRewards);
-        rewards = rewards.add(_rewards);
-        return stakePower;
+        uint256 delegatorsRewards = _rewards.sub(_validatorRewards);
```
atvanguard, line 46: Declaring this above line 41 and using it in line 43 might help readability.

test/helpers/deployer.js
```diff
@@ -412,10 +418,15 @@ class Deployer {
 
   async initializeChildChain(owner, options = { updateRegistry: true }) {
     // not mentioning the gas limit fails with "The contract code couldn't be stored, please check your gas limit." intermittently which is super weird
-    this.childChain = await contracts.ChildChain.new({ gas: 7500000 })
+    this.childChain = await contracts.ChildChain.new({ gas: 75000000 })
     await this.childChain.changeStateSyncerAddress(owner)
     if (!this.globalMatic) {
       // MRC20 comes as a genesis-contract at utils.ChildMaticTokenAddress
+      if (!utils.ChildMaticTokenAddress) {
```
atvanguard, line 425: Please explain this if block? `utils.ChildMaticTokenAddress` is set right?

test/helpers/deployer.js
```diff
@@ -412,10 +418,15 @@ class Deployer {
 
   async initializeChildChain(owner, options = { updateRegistry: true }) {
     // not mentioning the gas limit fails with "The contract code couldn't be stored, please check your gas limit." intermittently which is super weird
-    this.childChain = await contracts.ChildChain.new({ gas: 7500000 })
+    this.childChain = await contracts.ChildChain.new({ gas: 75000000 })
```
atvanguard, line 421: I understand that you need more gas for coverage. But for normal tests does bor have enough gas limit? If not we can use something `gas: process.env.SOLIDITY_COVERAGE ? 75000000 : 7500000`.

contracts/staking/validatorShare/ValidatorShare.sol
imyourm8: yeah i have this code in other branch "improve-naming-stake-manager" did same there.

test/helpers/deployer.js
imyourm8: Yes, it's a quick solution to re-deploy matic native token on the ganache instead during coverage. Will be removed when multiple providers will be supported.

test/helpers/deployer.js
imyourm8: Side effect of messing around, no need to increase at all i think.

test/helpers/utils.js
```diff
@@ -16,23 +16,36 @@ const BN = ethUtils.BN
...
   '0x0000000000000000000000000000000000001010'
+if (process.env.SOLIDITY_COVERAGE) {
+  ChildMaticTokenAddress = null
+}
 export const scalingFactor = web3.utils.toBN(10).pow(web3.utils.toBN(18))
 
 export function getSigs(wallets, votedata) {
-  wallets.sort((w1, w2) => {
+  // avoid any potential side effects
+  const copyWallets = [...wallets]
```
0xAshish, line 41: nice

test/units/staking/stakeManager/StakeManager.Staking.js
```diff
@@ -0,0 +1,786 @@
...
+
+      it('must emit UnstakeInit', async function() {
+        await expectEvent.inTransaction(this.receipt.tx, StakingInfo, 'UnstakeInit', {
+          amount: amounts.stakeAmount,
+          validatorId: this.validatorId,
+          user
+        })
+      })
+
+      // for some reason DummyERC20 doesn't have Transfer event
```
0xAshish, line 310: if you are using truffle log parser then give a try to use test token instead od using DummyErc20. Due to the same event `Transfer` in both erc721/erc20, I wasn't able to parse that log so removed in Dummy one.

imyourm8: I see, i'll get it back then

PR #236, opened 2020-05-06
[WIP] Validators staking txs nonces

engwarrior:
https://linear.app/issue/MAT-1109/sequence-for-ethereum-tx-order-issue

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>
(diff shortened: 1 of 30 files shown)


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
0xAshish, line 1: Why push this file?

contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -721,6 +742,7 @@ contract StakeManager is IStakeManager, StakeManagerStorage {
         logger.logStaked(
             signer,
             signerPubkey,
+            1,
```
imyourm8, line 745: 1 - here is validatorId not nonce, check StakingInfo logStaked arguments order.

contracts/staking/stakeManager/StakeManagerStorage.sol
```diff
@@ -55,6 +55,7 @@ contract StakeManagerStorage is ProxyStorage, Lockable, RootChainable {
         uint256 activationEpoch;
         uint256 deactivationEpoch;
         uint256 jailTime;
+        uint256 nonce;
```
0xAshish, line 58: Please don't add nonce into validator struct, keep separate mapping in staking info itself. Since all the required log functions have `validatorId` in it.

0xAshish: We don't want to make our validator struct heavy and this obviously makes more sense if we add it in stakingInfo/logger only.

0xAshish:
Closing this pr since changes are merged in https://github.com/maticnetwork/contracts/pull/243

PR #243, opened 2020-05-08
add validator nonce in staking info contract (mat 1109)

AyushKaul:
Only tests for staking and re-staking are added.

### DIFF
--- a/contracts/staking/StakingInfo.sol
+++ b/contracts/staking/StakingInfo.sol
@@ -23,6 +23,7 @@ contract IStakeManager {
     }
 
     mapping(uint256 => Validator) public validators;
+    mapping(uint256 => uint256) public validatorNonce;
     bytes32 public accountStateRoot;
     uint256 public activeAmount; // delegation amount from validator contract
     uint256 public validatorRewards;
@@ -45,6 +46,7 @@ contract StakingInfo {
     event Staked(
         address indexed signer,
         uint256 indexed validatorId,
+        uint256 indexed validatorNonce,
         uint256 indexed activationEpoch,
         uint256 amount,
         uint256 total,
@@ -53,6 +55,7 @@ contract StakingInfo {
     event Unstaked(
         address indexed user,
         uint256 indexed validatorId,
+        uint256 indexed validatorNonce,
         uint256 amount,
         uint256 total
     );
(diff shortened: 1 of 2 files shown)


contracts/staking/StakingInfo.sol
```diff
@@ -41,10 +41,12 @@ contract IStakeManager {
 
 contract StakingInfo {
     using SafeMath for uint256;
+    mapping(uint256 => uint256) public validatorNonce;
```
jdkanani, line 44: Shouldn't we put this in `StakeManagerStorage`?  Other than everything looks good. 

Also `validatorNonce` in `StakingInfo.sol` is okay 😄

contracts/staking/StakingInfo.sol
```diff
@@ -189,11 +197,13 @@ contract StakingInfo {
         emit Staked(
             signer,
             validatorId,
+            validatorNonce[validatorId].add(1),
             activationEpoch,
             amount,
             total,
             signerPubkey
         );
+        validatorNonce[validatorId] = validatorNonce[validatorId].add(1);
```
jdkanani, line 206: We should do it before `emit`, so we don't have to add twice. For all events.

contracts/staking/StakingInfo.sol
AyushKaul: `StakeManagerStorage` is used in many other places where the nonce will not make sense. @0xAshish has more context.

contracts/staking/StakingInfo.sol
AyushKaul: yes, I'll change it.

AyushKaul: done

test/staking/StakeManager.test.js
```diff
@@ -146,6 +145,8 @@ contract('StakeManager', async function (accounts) {
 
       logs[1].event.should.equal('Staked')
       logs[1].args.signerPubkey.toLowerCase().should.equal(userPubkey.toLowerCase())
+      // console.log(logs[1].args.nonce)
+      // assertBigNumberEquality(logs[1].args.nonce, 1)
```
jdkanani, line 149: Why removed?

contracts/staking/StakingInfo.sol
```diff
@@ -220,14 +232,16 @@ contract StakingInfo {
...
     }
 
     function logReStaked(uint256 validatorId, uint256 amount, uint256 total)
         public
         onlyStakeManager
-    {
-        emit ReStaked(validatorId, amount, total);
+    {   
+        validatorNonce[validatorId] = validatorNonce[validatorId].add(1);
+        emit ReStaked(validatorId, validatorNonce[validatorId], amount, total);
```
jdkanani, line 244: Only the following events need to emit with `nonce` (not any other event):

`Staked`, `StakeUpdate` , `SignerChange` , `UnstakeInit`

test/staking/StakeManager.test.js
AyushKaul: only comments

PR #246, opened 2020-05-11
Merge sidechannel to develop

0xAshish:


contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -356,7 +357,8 @@ contract StakeManager is IStakeManager, StakeManagerStorage {
         totalStaked = totalStaked.sub(amount);
 
         NFTContract.burn(validatorId);
-        signerToValidator[validators[validatorId].signer] = INCORRECT_VALIDATOR_ID;
+        signerToValidator[validators[validatorId]
+            .signer] = INCORRECT_VALIDATOR_ID;
```
imyourm8, line 361: Can you configure the linter so it doesn't break this line? It is probably max width parameter or something.

test/helpers/utils.js
```diff
@@ -53,6 +52,29 @@ export function getSigs(wallets, votedata) {
     .filter(d => d)
 }
 
+export function getSigsWithVotes(wallets, data, sigPrefix, maxYesVotes) {
+  wallets.sort((w1, w2) => {
```
imyourm8, line 56: Also make a copy of the wallets, last i time i was wondering why my input array was re-ordered.

0xAshish: done

contracts/staking/stakeManager/StakeManager.sol
0xAshish: what should be the max length?

PR #258, opened 2020-05-19
Mat 1462 sellVoucher reverts after slashing

imyourm8:


contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -147,6 +147,10 @@ contract ValidatorShare is IValidatorShare {
             _amount = _amount.sub(_rewards);
         }
 
+        if (_amount > activeAmount) {
+            _amount = activeAmount;
```
0xAshish, line 151: This is wrong.

imyourm8: Can you elaborate what is wrong and how would you do this?

0xAshish: bob stakes 200 and alice stakes 100, after which total pool 300 got slashed by 50% then we don't want to give bob 150 instead it should be 100(bob) and 50(Alice).

0xAshish:
Closing the PR since there are different set of changes required.

0xAshish:
Closing the PR since there are different sets of changes required.

PR #259, opened 2020-05-19
Mat 1463 ValidatorShare improvements

imyourm8:


contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -115,9 +115,12 @@ contract ValidatorShare is IValidatorShare {
     }
 
     function buyVoucher(uint256 _amount) public onlyWhenUnlocked {
-        uint256 share = _amount.mul(100).div(exchangeRate());
+        uint256 rate = exchangeRate();
+        uint256 share = _amount.mul(100).div(rate);
         require(share > 0, "Insufficient amount to buy share");
         require(delegators[msg.sender].share == 0, "Ongoing exit");
+        
+        _amount = _amount - (_amount % rate.mul(share).div(100));
```
0xAshish, line 123: safe math for subtracting `sub`

atvanguard, line 123: Nice trick 🎉

imyourm8: I don't think it is necessary here

PR #260, opened 2020-05-19
Mat 1461

imyourm8:


jdkanani:
@imyourm8 Please resolve conflict. Looks good otherwise.

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -212,13 +212,20 @@ contract ValidatorShare is IValidatorShare {
...
+            return 0;
+        }
+        
+        uint256 liquidRewards;
+        uint256 totalTokens = exchangeRate().mul(share).div(100);
+        if (totalTokens >= amountStaked[user]) {
+            liquidRewards = totalTokens.sub(amountStaked[user]);
+        }
+
+        return liquidRewards;
```
atvanguard, line 228: Less code -> easier on the eyes and saved gas.
```suggestion
        uint256 totalTokens = exchangeRate().mul(share).div(100);
        if (totalTokens >= amountStaked[user]) {
            return totalTokens.sub(amountStaked[user]);
        }
        return 0;
```

test/units/staking/ValidatorShare.test.js
```diff
@@ -695,4 +695,24 @@ contract('ValidatorShare', async function() {
       })
     })
   })
+
+  describe.only('getLiquidRewards', function() {
```
atvanguard, line 699: Remember to remove `.only` before merge and check for CI

contracts/staking/validatorShare/ValidatorShare.sol
imyourm8: I would argue that your version uses less gas. With early exit i don't read any of the storage variables. 32 bytes of memory expansion cost 1 gas, so in general my implementation is more efficient, i only need to read amountStaked[user] once.

atvanguard: Only case where your version requires less gas is when shares == 0. I personally like to optimize for the usual case, not an outlier.

PR #264, opened 2020-05-21
Make validator share upgradeable

atvanguard:


contracts/staking/validatorShare/IValidatorShare.sol
```diff
@@ -1,59 +1,9 @@
...
-        _transferOwnership(_stakeManager);
-    }
-
-    function updateRewards(uint256 _reward, uint256 _totalStake)
-        external
-        returns (uint256);
+import {ValidatorShareStorage} from "./ValidatorShareStorage.sol";
 
+
+contract IValidatorShare is ValidatorShareStorage {
```
atvanguard, line 6: Please don't inherit from `ValidatorShareStorage`. Add the method `activeAmount()` and don't inherit this in ValidatorShare.
This can possibly mess with the storage in ValidatorShare.

0xAshish: So many methods are required due to storage var reads.

PR #265, opened 2020-05-21
Update delegation contract address

0xAshish:


contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -389,6 +326,21 @@ contract StakeManager is IStakeManager, StakeManagerStorage {
         CHECKPOINT_REWARD = newReward;
     }
 
+    // Change delegation contract for a validator
+    // @note: Users must exit before this update or all funds may get lost
+    function updateContractAddress(uint256 validatorId, address newContractAddress) public onlyOwner {
+        require(
```
imyourm8, line 332: what if previous contract will require shares balance to be 0 ?

contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -230,27 +208,7 @@ contract StakeManager is IStakeManager, StakeManagerStorage {
         _unstake(validatorId, currentEpoch);
     }
 
-    // Housekeeping function. @todo remove later
-    // other than varibale you want to update give same values
-    function updateConstructor(
```
jdkanani, line 235: Shouldn't we keep this? Otherwise how we are going to change `_stakingLogger` or `_rootchain`?

contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -346,20 +297,6 @@ contract StakeManager is IStakeManager, StakeManagerStorage {
         logger.logClaimRewards(validatorId, amount, totalRewardsLiquidated);
     }
 
-    // returns valid validator for current epoch
-    function getCurrentValidatorSet() public view returns (uint256[] memory) {
```
jdkanani, line 350: Are there any other methods to get all validators?

contracts/staking/stakeManager/StakeManager.sol
jdkanani: @imyourm8 we shouldn't add that condition. So let's say in the worst condition, governance contracts can drain the balance and deploy contract and upgrade contract address if delegators can't withdraw funds.

contracts/staking/stakeManager/StakeManager.sol
0xAshish: This wasn't used by any contract and the external party can read the blockchain and using same as this function.

PR #278, opened 2020-05-29
 Audit fixes

0xAshish:


contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -81,11 +82,12 @@ contract ValidatorShare is ValidatorShareStorage {
...
 
-    function sellVoucher() public {
+    function sellVoucher(uint256 expectedExchangeRate) public {
         uint256 share = balanceOf(msg.sender);
         require(share > 0, "Zero balance");
-        uint256 _amount = exchangeRate().mul(share).div(EXCHANGE_RATE_PRECISION);
-
+        uint256 rate = exchangeRate();
+        uint256 _amount = rate.mul(share).div(EXCHANGE_RATE_PRECISION);
+        require(rate >= expectedExchangeRate, "More slippage then expectedExchangeRate");
```
jdkanani, line 90: We should move to `89`. We should do the calculation after the require condition.

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -217,12 +219,4 @@ contract ValidatorShare is ValidatorShareStorage {
         locked = true;
         return activeAmount;
     }
-
```
jdkanani, line 220: Please add test case for the same.

contracts/staking/validatorShare/ValidatorShare.sol
0xAshish: ah yes

contracts/staking/validatorShare/ValidatorShare.sol
0xAshish: yup adding

contracts/root/RootChain.sol
```diff
@@ -16,38 +16,19 @@ contract RootChain is RootChainStorage, IRootChain {
...
-    {
-        (
-            address proposer,
-            uint256 start,
-            uint256 end,
-            bytes32 rootHash,
-            bytes32 accountHash,
-            uint256 _borChainID
-        ) = abi.decode(
-            data,
```
AyushKaul, line 37: this looks more readable

0xAshish: this is due to max length of the line. We can change it in the config.
:D bad idea for this PR.

contracts/root/depositManager/DepositManager.sol
```diff
@@ -132,14 +105,13 @@ contract DepositManager is
...
     }
 
-    function depositERC20ForUser(address _token, address _user, uint256 _amount)
-        public
-    {
+    function depositERC20ForUser(
+        address _token,
+        address _user,
+        uint256 _amount
+    ) public {
```
jdkanani, line 112: This lint reversed!

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -59,8 +59,9 @@ contract ValidatorShare is ValidatorShareStorage {
                 : withdrawPool.mul(EXCHANGE_RATE_PRECISION).div(_withdrawShares);
     }
 
-    function buyVoucher(uint256 _amount) public onlyWhenUnlocked {
+    function buyVoucher(uint256 _amount, uint256 expectedExchangeRate) public onlyWhenUnlocked {
```
atvanguard, line 62: it's better to take `minShareAmount` as an input parameter, then
```
require(share >= minShareAmount, "Insufficient amount to buy share");
```

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -81,11 +82,12 @@ contract ValidatorShare is ValidatorShareStorage {
         logger.logStakeUpdate(validatorId);
     }
 
-    function sellVoucher() public {
+    function sellVoucher(uint256 expectedExchangeRate) public {
```
atvanguard, line 85: Same as above.

PR #279, opened 2020-05-30
dev: change buy/sell voucher signature

atvanguard:


scripts/stakeManagerTest.js
```diff
@@ -36,7 +36,7 @@ async function stake() {
   const _validator = await stakeManager.validators(validatorId)
   const validator = await ValidatorShare.at(_validator.contractAddress)
   console.log('approved, delegating now...')
-  result = await validator.buyVoucher(web3.utils.toWei('100'), await validator.exchangeRate())
+  result = await validator.buyVoucher(web3.utils.toWei('100'), 0)
```
jdkanani, line 39: We should add one more test case for equal condition.

atvanguard: Added. Please check.

PR #281, opened 2020-05-30
Set exit period = 1 week + increase leeway for processExits

atvanguard:


contracts/root/withdrawManager/WithdrawManagerStorage.sol
```diff
@@ -65,7 +66,7 @@ contract WithdrawManagerStorage is ProxyStorage, WithdrawManagerHeader {
     // uint32 constant internal ITERATION_GAS = 52000;
 
     // So putting an upper limit of 155000 + 52000 + leeway
-    uint32 internal constant ON_FINALIZE_GAS_LIMIT = 210000;
+    uint32 internal constant ON_FINALIZE_GAS_LIMIT = 300000;
```
0xAshish, line 69: We should keep this var updatable so that we can handle Istanbul like incidents.

0xAshish: or at least please don't make it `constant` so that we can upgrade proxy and add a function to update this.

PR #285, opened 2020-06-08
Unstructured storage proxy for validator share

imyourm8:


contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -1,24 +1,59 @@
...
-        uint256 _validatorId,
-        address _stakingLogger,
-        address _stakeManager
-    ) public Lockable(_stakeManager) {} // dummy constructor
+    // onlyOwner will prevent this contract from initializing, since it's owner is going to be 0x0 address
+    function initialize(uint256 _validatorId, address _stakingLogger, address _stakeManager) external initializer  {
+        validatorId = _validatorId;
+        stakingLogger = StakingInfo(_stakingLogger);
+        stakeManager = IStakeManager(_stakeManager);
+        _transferOwnership(_stakeManager);
```
jdkanani, line 52: Since `stakeManager` will be the owner for validator share proxy, only it can update the implementation for validator share, right? 

Shouldn't we allow governance to update proxy?

contracts/staking/validatorShare/ValidatorShareFactory.sol
```diff
@@ -1,14 +1,28 @@
...
+        address proxyAddr = address(proxy);
+        (bool success, bytes memory data) = proxyAddr.call.gas(gasleft())(
+            abi.encodeWithSelector(
+                ValidatorShare(proxyAddr).initialize.selector, 
+                validatorId, 
+                loggerAddress, 
+                msg.sender
+            )
+        );
+        require(success, string(data));
```
jdkanani, line 24: Why need to use `initialize` instead of using constructor? Because of how the proxy works? In that case, `registry` will be on proxy instead of implementation, right?

contracts/staking/validatorShare/ValidatorShare.sol
imyourm8: Ownership transfer you highlighted is owner of the ValidatorShare implementation. Implementation lives in the Registry, to update every validator share contract we just need to swap address in the registry itself

contracts/staking/validatorShare/ValidatorShareFactory.sol
imyourm8: Initialize is required because constructor already was invoked when we deploy implementation. We use proxy's storage, but layout lives in the implementation now. Hence, proxy's ctor is empty, storage also empty.

Before proxy inherited storage layout too which enables constructor with default storage values.

Yes, registry is default implementation in the proxy, as it was before. Then i obtain actual implementation from the registry and return it for delegatecall.

PR #292, opened 2020-06-16
binary compiler

imyourm8:
Without optimization enabled StakeManager contract size is at ~38.6 kb out of 24kb max.Which suggests that we either need to heavily refactor it or take a risk and optimize contracts.

### DIFF
--- a/truffle-config.js
+++ b/truffle-config.js
@@ -77,7 +77,9 @@ module.exports = {
   },
   compilers: {
     solc: {
-      version: '0.5.11',
+      version: '0.5.17',
+      docker: true,
+      parser: 'solcjs',
       settings: {
         optimizer: {
           enabled: true,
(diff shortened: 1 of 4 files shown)


imyourm8:
1st possible thing to refactor is use **error codes** instead of full text messages (like http protocol).

truffle-config.js
```diff
@@ -77,7 +77,9 @@ module.exports = {
   },
   compilers: {
     solc: {
-      version: '0.5.11',
+      version: '0.5.17',
+      docker: true,
+      parser: 'solcjs',
```
jdkanani, line 82: Are we still using solcjs? shouldn't we use sold binary?

imyourm8: this is only for parsing, truffle says it is faster than binary.

PR #295, opened 2020-06-17
chg: use github ci instead of travis

jdkanani:


imyourm8:
I am not expert, for me it looks good.

PR #307, opened 2020-07-02
set url from env variables

mankenavenkatesh:


moonwalker-migrations/deployWorker.js
```diff
@@ -3,9 +3,16 @@ const HDWalletProvider = require('truffle-hdwallet-provider')
...
-const url = `https://goerli.infura.io/v3/${API_KEY}`
-// const url = `http://localhost:8545`
+// set url
+if (NETWORK_URL && API_KEY) {
+  url = `${NETWORK_URL}/${API_KEY}`
+} else if (NETWORK_URL) {
+  url = `${NETWORK_URL}`
+} else {
+  url = `http://localhost:8545`
+}
```
atvanguard, line 15:
```suggestion
// set url
let url = `http://localhost:8545`
if (NETWORK_URL) {
  url = `${NETWORK_URL}`
}
```

atvanguard: Let's keep only network url then and remove API_KEY.

PR #317, opened 2020-07-29
change Exitor to exit nft owner (MAT-1996)

AyushKaul:


test/integration/root/ExitScenarios.test.js
```diff
@@ -15,7 +15,7 @@ const web3Child = utils.web3Child
 chai.use(chaiAsPromised).should()
 let contracts, childContracts, rootERC20, childErc20, statefulUtils
 
-contract('Misc Predicates tests @skip-on-coverage', async function(accounts) {
+contract.only('Misc Predicates tests @skip-on-coverage', async function(accounts) {
```
imyourm8, line 18: Remove `only`

contracts/root/withdrawManager/WithdrawManager.sol
```diff
@@ -233,11 +233,9 @@ contract WithdrawManager is WithdrawManagerStorage, IWithdrawManager {
             exitQueue.delMin();
             // If the exitNft was deleted as a result of a challenge, skip processing this exit
             if (!exitNft.exists(exitId)) continue;
-
+            address exitor = exitNft.ownerOf(exitId);
+            exits[exitId].owner = exitor;
```
jdkanani, line 237: Do we need to change owner in exists?  @atvanguard @AyushKaul

AyushKaul: its used while encoding so changed it.
cc: @atvanguard

test/integration/root/ExitScenarios.test.js
AyushKaul: trying to run tests locally first

AyushKaul:
added test to transfer exit nft before processing the exit.

contracts/root/withdrawManager/WithdrawManager.sol
jdkanani: Do we need it here though?

PR #322, opened 2020-08-11
Partial unstake

imyourm8:


imyourm8:
I started working on develop branch, maybe i shouldn't merge into release with it.

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -180,29 +203,33 @@ contract ValidatorShare is IValidatorShare, ERC20NonTransferable, OwnableLockabl
         return _amount;
     }
 
-    function sellVoucher(uint256 _minClaimAmount) public {
-        uint256 shares = balanceOf(msg.sender);
-        require(shares > 0, "Zero balance");
+    function sellVoucher(uint256 claimAmount) public {
```
jdkanani, line 206: Shouldn't we put `minimumShares` with `claimAmount`?

imyourm8: Depends how we want it to be. I can do it in any way.

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -161,14 +185,13 @@ contract ValidatorShare is IValidatorShare, ERC20NonTransferable, OwnableLockabl
...
-        require(delegators[msg.sender].shares == 0, "Ongoing exit");
+        require(unbonds[msg.sender].shares == 0, "Ongoing exit");
 
         _mint(msg.sender, shares);
 
         _amount = _amount.sub(_amount % rate.mul(shares).div(EXCHANGE_RATE_PRECISION));
 
-        totalStake = totalStake.add(_amount);
-        amountStaked[msg.sender] = amountStaked[msg.sender].add(_amount);
+        // totalStake = totalStake.add(_amount);
```
jdkanani, line 194: Why `totalStake` removed? How we are going to get `totalStake` for this validator?

contracts/staking/validatorShare/ValidatorShare.sol
imyourm8: Actually, minimum shares doesn't make sense, because shares doesn't get slashed at all. Only amount of tokens can be lowered.

jdkanani: but claimAmount to shares can be changed. So `shares` changes up or down just before calling `sellVoucher`?

PR #331, opened 2020-09-01
Fix merkle patricia library

arthcp:
- Fixes merkle patricia proof verification in solidity for the case where there is an extension node in the receipt node's path. `pathPtr` got incremented using return value of `_nibblesToTraverse`. This happened before the extension node check and it resulted in verification failure.
As a fix, storing the return value of `_nibblesToTraverse` in a variable and using it when needed.

- Fix receipt status in `getReceiptBytes` test helper function.

- Fix branch mask encoding. HP encoding is required instead of RLP encoding. Since path will always be even number of nibbles, just prepending `00` is enough.


arthcp:
Similar fix done in https://github.com/maticnetwork/pos-portal/pull/42

contracts/common/lib/MerklePatriciaProof.sol
```diff
@@ -69,12 +69,12 @@ library MerklePatriciaProof {
...
                 pathPtr += 1;
             } else if (currentNodeList.length == 2) {
-                pathPtr += _nibblesToTraverse(
+                uint256 traversed = _nibblesToTraverse(
                     RLPReader.toBytes(currentNodeList[0]),
                     path,
                     pathPtr
                 );
-                if (pathPtr == path.length) {
+                if (pathPtr + traversed == path.length) {
```
AyushKaul, line 77: @arthcp is safe math required here?

jdkanani: No need I guess.

AyushKaul (approved):

imyourm8 (approved):

PR #332, opened 2020-09-08
Matic Layer 2 Security Hackathon

levalleux-ludo:
Please see my 3 references to reports in HackerOne


imyourm8:
It should be an issue.

PR #339, opened 2020-12-25
Make validator shares transferable

web3jim:
This change makes validator shares transferable. See [discussion](https://forum.matic.network/t/make-validator-shares-transferable-to-allow-for-key-rotation/563).


jdkanani:
Thanks @web3jim, for the PR.

After discussing with @imyourm8, we realized it is not as easy as transferring shares since we also need to calculate rewards and transfer for `from` and `to` before the actual transfer.

@imyourm8 will take work to enable trasfers and fix test cases along the way.

imyourm8:
Opened different PR #340.

PR #341, opened 2020-12-28
new: independent shares exits

imyourm8:


contracts/staking/StakingInfo.sol
```diff
@@ -466,9 +467,10 @@ contract StakingInfo is Ownable {
         uint256 validatorId,
         address user,
         uint256 amount,
-        uint256 tokens
+        uint256 tokens,
+        uint256 burnId
     ) public onlyValidatorContract(validatorId) {
-        emit ShareBurned(validatorId, user, amount, tokens);
+        emit ShareBurned(validatorId, user, amount, tokens, burnId);
```
jdkanani, line 473: Should we also change the name of the event? Otherwise indexing it would be hard. 
Something like `SharedBurnedWithBurnId`?

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -236,6 +232,60 @@ contract ValidatorShare is IValidatorShare, ERC20NonTransferable, OwnableLockabl
...
+        _withdrawAndTransferReward(msg.sender);
+
+        _burn(msg.sender, shares);
+        stakeManager.updateValidatorState(validatorId, -int256(claimAmount));
+
+        uint256 _withdrawPoolShare = claimAmount.mul(precision).div(withdrawExchangeRate());
+        withdrawPool = withdrawPool.add(claimAmount);
+        withdrawShares = withdrawShares.add(_withdrawPoolShare);
+
+        uint256 unbondNonce = unbondNonces[msg.sender].add(1);
```
jdkanani, line 258: Should we update after we set it by`unbonds_new[msg.sender][unbondNonce] = unbond` ? So it can start with `0`? or it doesn't matter I guess.

contracts/staking/validatorShare/ValidatorShare.sol
```diff
@@ -236,6 +232,60 @@ contract ValidatorShare is IValidatorShare, ERC20NonTransferable, OwnableLockabl
...
+        );
+
+        uint256 _amount = withdrawExchangeRate().mul(shares).div(_getRatePrecision());
+        withdrawShares = withdrawShares.sub(shares);
+        withdrawPool = withdrawPool.sub(_amount);
+
+        require(stakeManager.transferFunds(validatorId, _amount, msg.sender), "Insufficent rewards");
+        stakingLogger.logDelegatorUnstaked(validatorId, msg.sender, _amount);
+        delete unbonds_new[msg.sender][unbondNonce];
+    }
```
jdkanani, line 287: Shouldn't we reuse this code with `unstakeClaimTokens` by having `_unstakeClaimTokens(DelegatorUnbond  memory unbond)`? Just a small code structuring.

jdkanani (changes requested):

contracts/staking/StakingInfo.sol
imyourm8: I am not sure, if it is better to have a separate event, I will do that.

jdkanani: On bad thing, I realized that we don't have a proxy in stakingInfo!

jdkanani (approved):

PR #348, opened 2021-02-25
Fixes v2

imyourm8:


contracts/staking/stakeManager/StakeManagerExtension.sol
```diff
@@ -128,13 +128,18 @@ contract StakeManagerExtension is StakeManagerStorage, Initializable, StakeManag
...
             ValidatorShare contractAddress = ValidatorShare(validators[i].contractAddress);
-            if (contractAddress != ValidatorShare(0)) {
-                validators[i].delegatorsReward = contractAddress.validatorRewards_deprecated().add(INITIALIZED_AMOUNT);
+            if (contractAddress != ValidatorShare(0) && !validators[i].migrated) {
+                validators[i].migrated = true;
+                // move validator rewards out from ValidatorShare contract
+                validators[i].rewards = contractAddress.validatorRewards_deprecated().add(INITIALIZED_AMOUNT);
                 validators[i].delegatedAmount = contractAddress.activeAmount();
                 validators[i].commissionRate = contractAddress.commissionRate_deprecated();
+                validators[i].delegatorsReward = INITIALIZED_AMOUNT;
```
jdkanani, line 137: Do we need it here? We are already doing it after this if-else.

contracts/staking/stakeManager/StakeManagerExtension.sol
```diff
@@ -128,13 +128,18 @@ contract StakeManagerExtension is StakeManagerStorage, Initializable, StakeManag
     function migrateValidatorsData(uint256 validatorIdFrom, uint256 validatorIdTo) external {       
         for (uint256 i = validatorIdFrom; i < validatorIdTo; ++i) {
             ValidatorShare contractAddress = ValidatorShare(validators[i].contractAddress);
-            if (contractAddress != ValidatorShare(0)) {
-                validators[i].delegatorsReward = contractAddress.validatorRewards_deprecated().add(INITIALIZED_AMOUNT);
+            if (contractAddress != ValidatorShare(0) && !validators[i].migrated) {
+                validators[i].migrated = true;
```
jdkanani, line 132: Do we need it because we will be running migration only once?

contracts/staking/stakeManager/StakeManager.sol
```diff
@@ -945,7 +945,9 @@ contract StakeManager is
             }
         }
 
-        validators[validatorId].initialRewardPerStake = newRewardPerStake;
+        if (newRewardPerStake > initialRewardPerStake) {
```
jdkanani, line 948: This is in case of slashing or normal scenario also?

imyourm8: Rewards should only grow, slashing works by slashing shares.

contracts/staking/stakeManager/StakeManagerExtension.sol
imyourm8: Not necessary but was thinking to prevent any potential breaking

contracts/staking/stakeManager/StakeManagerExtension.sol
imyourm8: No need, probably forgot to remove it from if clause

jdkanani (approved):

PR #355, opened 2021-04-23
Checkpoints optimization v2

imyourm8:


imyourm8:
Removed 3 SLOAD instructions from the loop inside `checkSignatures`. It results in 600k gas improvement over the previous implementation.

PR #375, opened 2021-08-19
update npm module

zorancuc:


migrations/5_deploy_child_contracts.js
```diff
@@ -7,9 +7,11 @@ const ChildChain = artifacts.require('ChildChain')
 const MRC20 = artifacts.require('MRC20')
 
 module.exports = async function(deployer, network, accounts) {
-  if (deployer.network !== 'bor') {
-    return
-  }
+  // if (deployer.network !== 'bor') {
```
jdkanani, line 10: This comment is intentional?

jdkanani (changes requested):

migrations/5_deploy_child_contracts.js
zorancuc: Yeah. We wanted to deploy contracts on bordevnet. In this case deployer.network must be "bordevnet".
The commented code will return the function and migration code will not run. So I commented it.

If this commented code is needed part, then we can add one more condition.
if (deployer.network !== 'bor' && deployer.network !== 'bordevnet') {
   return
}

truffle-config.js
```diff
@@ -32,6 +32,15 @@ module.exports = {
       network_id: '*', // match any network
       gasPrice: '0'
     },
+    bordevnet: {
+      provider: () =>
+        new HDWalletProvider(
+          MNEMONIC,
+          `http://3.94.19.25:9545`
```
imyourm8, line 39: what is it for? Why new network is used?

zorancuc: It was used for safemath deployment test.

jdkanani (approved):

PR #396, opened 2022-01-26
Fixes

jdkanani:
- Fixes a bug where signers list goes out of sync when called update signer after unstake
- Fixes a bug where total stake was reducing twice while doing delegation migration after validator unstake


Fqc301033 (commented):
1

Fqc301033 (commented):
S

Abeanna (approved):
Ok

PR #424, opened 2022-10-13
dependency update to allow node 16

gatsbyz:
moving around the dev dependencies to prod. this is to enable  matic-cli to run `npm install --omit=dev` without having to install unnecessary dev dependencies. i suspect there is more cleaning to do but this will redundantly enable faster project installation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,17 +43,12 @@
   },
   "devDependencies": {
     "@openzeppelin/test-helpers": "^0.5.5",
-    "babel-cli": "^6.26.0",
     "babel-eslint": "10.0.1",
-    "babel-polyfill": "^6.26.0",
-    "babel-preset-env": "1.7.0",
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-stage-2": "^6.24.1",
     "babel-preset-stage-3": "^6.24.1",
     "bip39": "^2.5.0",
     "bluebird": "^3.5.5",
-    "chai": "4.2.0",
-    "chai-as-promised": "^7.1.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -62,20 +58,24 @@
     "eslint-plugin-standard": "4.0.0",
     "eth-gas-reporter": "0.1.12",
     "eth-sig-util": "^2.1.1",
-    "ethereumjs-wallet": "0.6.2",
+    "ethereumjs-wallet": "0.6.3",
     "ethers": "4.0.13",
-    "moonwalker": "github:atvanguard/moonwalker#4c4592e2d6c1744fbcd9452cf0f00e9beb236b21",
+    "moonwalker": "github:gatsbyz/moonwalker",
```
praetoriansentry, line 63: If we're going to do something like this, it should be moved into the `maticnetwork` organization.

truffle-config.js
```diff
@@ -76,7 +76,7 @@ module.exports = {
   compilers: {
     solc: {
       version: '0.5.17',
-      docker: true,
+      docker: false,
```
praetoriansentry, line 79: Curious about this one. Is this required?

gatsbyz: config docs: https://trufflesuite.com/docs/truffle/reference/configuration/#solc
```
docker: <boolean>, // Use a version obtained through docker
```

with the bumped node version i am getting errors like this for some reason `Error: Error: spawnSync /bin/sh ENOBUFS`. turning this docker version off fixed it (as stackoverflow suggested). i am not sure if this is the right way to go.

